PR #35, opened 2016-03-18
Should fix #28

### LINKED ISSUE #28: Add a mobile view
It would be a nice feature if there was a mobile template for when the site if viewed from a phone.

shiitake:


tidusjar:
One of the test failed, it shouldn't really be a test if it can fail by exceeding an api limit. so dont worry about the failed build.

PR #39, opened 2016-03-19
Dev

shiitake:


PlexRequests.Store/SqlTables.sql
```diff
@@ -35,4 +35,25 @@ CREATE TABLE IF NOT EXISTS Log
     Message								varchar(100) NOT NULL,
     CallSite							varchar(100) NOT NULL,
     Exception							varchar(100) NOT NULL
+);
+
+CREATE TABLE IF NOT EXISTS Requested
```
tidusjar: Can we take this out please :)

PlexRequests.UI/Modules/UserLoginModule.cs
```diff
@@ -98,9 +100,17 @@ private Response LoginUser()
...
+                        authenticated = true;
+                    }
+                    else
+                    {
+                        authenticated = CheckIfUserIsInPlexFriends(username, settings.PlexAuthToken);
+                        Log.Debug("Friends list result = {0}", authenticated);
+                    }
                 }
             }
             else if(settings.UserAuthentication) // Check against the users in Plex
```
tidusjar: We also need to check if it's an owner in this case, where we don't have the password enabled but we do want to authenticate

tidusjar:
We should add some unit tests to cover this:
- Login requires username & password, user is not in friends but is the owner
- Login requires username only, user is not in friends but is the owner

Issue #62, opened 2016-03-21
Feature Request: Add the ability to request the first season


tidusjar:
https://www.reddit.com/r/PleX/comments/4b3thc/plex_requestsnet_release_140/d17to5k?context=3

shiitake:
Looking into this enhancement.

Issue #68, opened 2016-03-21
Add ability to choose type of monitoring when adding show to Sonarr


mattg1000:
Currently in version 1.4.1 when a new show is added to Sonarr by choosing the Last Season only option, it add the show but only monitors future episodes.  This causes Sonarr to never download the full season.

Please see screenshot below: 

![capture](https://cloud.githubusercontent.com/assets/1860530/13935474/770857cc-ef85-11e5-9941-063b0652e214.PNG)

tidusjar:
Yeah that is the correct behaviour. I think it's just mislabelled. I followed what the meteor version of plex requests did. 

In the next release I will be changing that dropdown to something like the following:

- First Season
- All Sesons
- Future Episodes. 

Then shortly after I will implement the ability to choose a specific season (another dropdown option). 

What do you think?

mattg1000:
What I'm noticing is that when I choose Latest Season in Plex Requests, it does not set the episodes to be monitored as is referenced in the screenshot above.

When I manually add a new show in Sonarr and choose Latest Season:
![image](https://cloud.githubusercontent.com/assets/1860530/13938305/95f4695e-ef9a-11e5-9bc4-6a8378e7b0f8.png)

I get the following results for the episodes in that season:
![image](https://cloud.githubusercontent.com/assets/1860530/13938313/ac4c0df6-ef9a-11e5-9c89-8efec2b3613a.png)

I was wanting to see the same behavior when a new show is added from Plex Requests.  Thanks for your work on this!

tidusjar:
Oh really. My understanding is incorrect.
I will fix this then. Thanks.

tidusjar:
This is related to #62

PR #83, opened 2016-03-24
Fixes issue #62 and #68

shiitake:


PlexRequests.Api/SonarrApi.cs
```diff
@@ -64,24 +64,28 @@ public SonarrAddSeries AddSeries(int tvdbId, string title, int qualityId, bool s
...
+            //}
+            //else
+            //{
+            //    options.addOptions = new AddOptions
+            //    {
+            //        ignoreEpisodesWithFiles = false,
+            //        ignoreEpisodesWithoutFiles = false,
+            //        searchForMissingEpisodes = true
+            //    };
+            //}
```
shiitake: We can probably just yank this code but wanted you to verify. 

I did a lot of back and forth in Postman to see if this would have an impact on the settings in sonarr and with the other changes we're making I don't think it needs to be here.

Issue #86, opened 2016-03-25
Show the item name in the "Success" or "Already in Plex" notifications


Auwen:
Enhancement to just say something like "[Movie_Title] ([Year]) successfully added!" instead of just "Successfully Added", etc.

This way, when adding multiple items quickly, and one fails, you can easily tell which one.

shiitake:
Good call!

PR #102, opened 2016-03-26
show the movie/show title when requesting.

Drewster727:
modify the json response to include a message that contains the title


Drewster727:
#86 -- I'm going to add to this so the failures show the name. also need to add the year of the show/movie to the message.

PR #110, opened 2016-03-27
misc changes (sorting, ui, requesting, testing, other)

Drewster727:
The request page sorting was not working at all. Also added sorting by release date.


tidusjar:
It's broken the animation when I click on the other tab and try and sort?

Drewster727:
I will take another look at refactoring that part this evening so it's more independent of specific notification types @tidusjar

Drewster727:
@tidusjar should be good to go on the notifications

tidusjar:
The build is also complaining about one of the unit tests being broken around the notification service.

tidusjar:
In a first time setup for a user, I tried testing the email notification but it did not work. This is because the email has never been enabled so it has not subscribed to the notification service. 

Maybe on the test we need to subscribe and then unsubscribe after?

PlexRequests.UI/Content/search.js
```diff
@@ -7,6 +7,8 @@
 
 var searchSource = $("#search-template").html();
 var searchTemplate = Handlebars.compile(searchSource);
+var noResultsHtml = "<div class='no-search-results'>" +
```
tidusjar: :+1: I like it!

PlexRequests.Services/Interfaces/INotification.cs
```diff
@@ -35,5 +36,7 @@ public interface INotification
         string NotificationName { get; }
 
         Task NotifyAsync(NotificationModel model);
+
+        Task NotifyAsync(NotificationModel model, Settings settings);
```
tidusjar: I still don't understand why we need to modify the interface for the settings?

Why can't we just use the `ISettingsService<T>` in the constructor (DI) and then call `GetSettings()`?

PlexRequests.UI/Views/Admin/Settings.cshtml
```diff
@@ -83,14 +83,38 @@
...
                 <div>
+                    <input type="text" class="form-control-custom form-control " id="NoApprovalUsers" name="NoApprovalUsers" placeholder="e.g. John, Bobby" value="@Model.NoApprovalUsers">
+                </div>
+            </div>
+
+            <div class="form-group">
+                <div class="checkbox">
                     <label>
-                        <input type="number" id="WeeklyRequestLimit" name="WeeklyRequestLimit" class="form-control form-control-custom " value="@Model.WeeklyRequestLimit">
+                        @if (Model.UsersCanViewOnlyOwnRequests)
```
tidusjar: Put this above the user list where all the other checkboxes are?

PlexRequests.UI/Views/Admin/Settings.cshtml
```diff
@@ -83,14 +83,38 @@
...
+                        {
+                            <input type="checkbox" id="UsersCanViewOnlyOwnRequests" name="UsersCanViewOnlyOwnRequests" checked="checked"><text>Users can view their own requests only</text>
+                        }
+                        else
+                        {
+                            <input type="checkbox" id="UsersCanViewOnlyOwnRequests" name="UsersCanViewOnlyOwnRequests"><text>Users can view their own requests only</text>
```
tidusjar: >Users can only view their requests

tidusjar:
An overall comment, I really like what you have done.

Some small notes:

- The feature where we are only allowing the users to see their own requests, if they request something that another user has requested it's not obvious. I suggest we possibly modify the `RequestedModel` to include a list of users that have requested that movie/tv show. So even know it's been requested, the 2nd user doesn't know (and wont get the error saying it's already been requested) and also can now see it  in the requested page(if you know what I mean).

- Not sure if I have mentioed this already but the 'Approve All' on the requests page can probably go now we have the 'Approve Movies' & 'Approve TV Shows'.

- The 'Approve Movies and TV Shows would look better in my opinion over to the right near the filters


Other than that and the comments above it's pretty cool! :D

PlexRequests.Services/Interfaces/INotification.cs
Drewster727: Discussed in gitter chat -- needed for passing in test settings before a save is actually committed.

tidusjar:
I got the following when loading the requests page as an admin:

PlexRequests.UI/Modules/RequestsModule.cs
```diff
@@ -84,7 +84,7 @@ private Response GetMovies()
             var dbMovies = Service.GetAll().Where(x => x.Type == RequestType.Movie);
             if (settings.UsersCanViewOnlyOwnRequests && !isAdmin)
             {
-                dbMovies = dbMovies.Where(x => x.RequestedBy.Equals(Session[SessionKeys.UsernameKey].ToString(), StringComparison.OrdinalIgnoreCase));
+                dbMovies = dbMovies.Where(x => x.UserHasRequested(Session[SessionKeys.UsernameKey].ToString()));
```
tidusjar: Since we are using `Session[SessionKeys.UsernameKey]` a few times in this method, we should store it in it's own local variable so we don't have to keep querying the session.

PlexRequests.UI/Modules/SearchModule.cs
```diff
@@ -179,11 +179,20 @@ private Response RequestMovie(int movieId)
             Log.Trace(movieInfo.DumpJson);
             //#if !DEBUG
 
+            var settings = PrService.GetSettings();
+
+            // check if the movie has already been requested
             Log.Info("Requesting movie with id {0}", movieId);
-            if (RequestService.CheckRequest(movieId))
+            var existingRequest = RequestService.CheckRequest(movieId);
+            if (existingRequest != null)
```
tidusjar: Might want to check if the `UserHasRequested` is also not null?

Drewster727: I'm doing that check right below. I needed the logic to be this way to return the proper message.

PlexRequests.UI/Modules/RequestsModule.cs
Drewster727: Done

PR #130, opened 2016-04-02
Sick rage async issue

shiitake:
Finally got the async issues sorted out. 

Also I've done some refactoring adding the SickRage base class. 

As usual any and all feedback is welcome. :)


tidusjar:
Looks good!

PR #192, opened 2016-04-29
Added Released property to RequestViewModel. Added Released filter to…

origamirobot:
… Requests page and hooked it up to the handlebars template. This is in relation to the feature request [here](https://github.com/tidusjar/PlexRequests.Net/issues/160). Check boxes havent been implemented like the request asks, but it's not possible to filter on released items in the list.


tidusjar:
Oh wow! Excellent. I shall take a look at it in a bit! :D

PR #269, opened 2016-05-26
Use HTTPS for accessing the poster images

seancallinan:
When serving PlexRequests.Net using a reverse proxy that uses HTTPS, poster images make the requests page 'insecure' because they are loading from an HTTP host. This host also supports HTTPS, so it should use HTTPS instead.


tidusjar:
Thanks for this! 

The only problem is that you have made this change in the master branch. We only checkin to master when we release, so master is all the current production code. Can you please make these changes in the dev branch.

There is another area this needs to be changed.

https://github.com/tidusjar/PlexRequests.Net/blob/master/PlexRequests.UI/Views/Search/Index.cshtml#L147

seancallinan:
Will do :)

PR #280, opened 2016-05-27
Use HTTPS for the poster images

seancallinan:
Use HTTPS for poster images, this removes the mixed-content warnings when serving the application via an HTTPS reverse proxy


tidusjar:
Looks good! Thanks a lot! :+1:

PR #531, opened 2016-09-08
Fix for when you have the base option set and trying to run the wizard (ie.. reverse proxy)

onedr0p:


tidusjar:
Awesome, thanks!

PR #535, opened 2016-09-13
The beginning of grabbing Sonarr root folders and approving requests with sepecific root folders

onedr0p:
Here's a brief rundown of what I did and what is left to do.

In the Admin Setting for Sonarr:
- Removed Folder Path textbox
- Added button to grab root folders
- Added drop down with Root folders
- The code will send an API request to Sonarr and grab the folders when you hit the 'Get Root Folders' button https://github.com/Sonarr/Sonarr/wiki/Rootfolder
- Updated some logic thru-out the program to accept a rootFolderId (like how you handled qualityId)

What's left to do:
- Figure out why when you save it won't update the cache with the root folders grabbed.
- It won't populate the root folders from cache when you save and reload the Admin Settings > Sonarr
- Implement Approving requests with selecting a root folder.

### DIFF
--- a/PlexRequests.Api.Models/Sonarr/SonarrRootFolder.cs
+++ b/PlexRequests.Api.Models/Sonarr/SonarrRootFolder.cs
@@ -1,0 +1,38 @@
+﻿#region Copyright
+// /************************************************************************
+//    Copyright (c) 2016 Jamie Rees
+//    File: SonarrProfile.cs
+//    Created By: Jamie Rees
+//   
+//    Permission is hereby granted, free of charge, to any person obtaining
+//    a copy of this software and associated documentation files (the
+//    "Software"), to deal in the Software without restriction, including
+//    without limitation the rights to use, copy, modify, merge, publish,
+//    distribute, sublicense, and/or sell copies of the Software, and to
+//    permit persons to whom the Software is furnished to do so, subject to
+//    the following conditions:
+//   
+//    The above copyright notice and this permission notice shall be
+//    included in all copies or substantial portions of the Software.
+//   
+//    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+//    EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+//    MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+//    NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+//    LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+//    OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+//    WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+//  ************************************************************************/
+#endregion
+
+using System.Collections.Generic;
+
+namespace PlexRequests.Api.Models.Sonarr
+{
+    public class SonarrRootFolder
+    {
+        public int id { get; set; }
+        public string path { get; set; }
+        public long freespace { get; set; }
+    }
+}
--- a/PlexRequests.UI/Models/RootFolderModel.cs
+++ b/PlexRequests.UI/Models/RootFolderModel.cs
@@ -1,0 +1,10 @@
+﻿namespace PlexRequests.UI.Models
+{
+    public class RootFolderModel
+    {
+        public string Id { get; set; }
+        public string Path { get; set; }
+        public long FreeSpace { get; set; }
+    }
+}
+


PlexRequests.UI/Modules/RequestsBetaModule.cs
```diff
@@ -186,6 +186,37 @@ private Negotiator LoadRequests()
 
             }
 
+            IEnumerable<RootFolderModel> rootFolders = new List<RootFolderModel>();
```
tidusjar: This module is not being used yet just a FYI, there is no code being used in this class

onedr0p: I figured as much, I just copied what I wrote in the RequestModule.cs :)

PlexRequests.UI/Modules/AdminModule.cs
```diff
@@ -470,6 +471,19 @@ private Response GetSonarrQualityProfiles()
             return Response.AsJson(profiles);
         }
 
+        private Response GetSonarrRootFolders()
+        {
+            var settings = this.Bind<SonarrSettings>();
```
tidusjar: Instead of doing the validation in the Javascript, you can check if the binding is valid. There are a bunch of validators e.g. SonarrSettingsValidator (If you look at some of the POST requests you should be able to see how we validate, and when there is invalid data we return the invalid message)

PlexRequests.UI/Views/Admin/Sonarr.cshtml
```diff
@@ -222,13 +264,52 @@
...
-        				$('#getSpinner').attr("class", "fa fa-times");
+                    $('#getSpinner').attr("class", "fa fa-times");
+                    generateNotify("Something went wrong!", "danger");
+                }
+            });
+        });
+
+        $('#getRootFolders').click(function (e) {
+
+            $('#getRootFolderSpinner').attr("class", "fa fa-spinner fa-spin");
```
tidusjar: See my comment in the Module, we can move the validation using Nancy's Validate stuff (see here https://github.com/NancyFx/Nancy/wiki/Nancy-and-Validation)

tidusjar:
Looks good so far!

So the caching is not working for you?

PlexRequests.UI/Modules/RequestsBetaModule.cs
tidusjar: @onedr0p Ok that's fine :)

onedr0p:
Nope, I hit the button to grab the root folders, click save and reload the page and they are wiped out. I am clueless as to why :/

Also after I added the validation stuff in SonarrSettingsValidator it won't accept you even hitting Save. Sorry my C# is very bad lol

tidusjar:
Hey that's fine. I can take a look tonight to see if I can find out what is going wrong.

> Also after I added the validation stuff in SonarrSettingsValidator it won't accept you even hitting Save
What wouldn't accept? Not sure I understand.

Your C# is not bad! You've been able to read and understand what is going on and there is some confusing areas so that's execllent!

onedr0p:
I have a strong Linux background, and I recently started a position as a VB.net web developer and I slowly getting up to par with the Microsoft environment.

tidusjar:
OK, so just looked at your code, you are storing the root folders into the cache (Default cache expiry is 20 minutes). When we save the sonarr settings [Here](https://github.com/onedr0p/PlexRequests.Net/blob/master/PlexRequests.UI/Modules/AdminModule.cs#L409) we are binding the results from the `<form>` being submitted and binding it to the model, in this case `SonarrSettings`. 

Now `SonarrSettings` does not contain a definition for the root folders, so they are not being saved into the database.
So when reloading the page, we query the database for the SonarrSettings and the root folders do not exist.

So what we need to do is add a new property onto the `SonarrSettings` something like a `string[]` to store those root folders (You already have it in the cache, so just get it from the cache and assign it to the new `string[]` property 👍 ).

If you need any help please ask 😄

PlexRequests.UI/Views/Admin/Sonarr.cshtml
```diff
@@ -157,11 +166,44 @@
...
+            if (!rootFolderSelected) {
+                return;
+            }
+            var $form = $("#mainForm");
+            $.ajax({
+                type: $form.prop("method"),
+                data: $form.serialize(),
+                url: "sonarrrootfolders",
+                dataType: "json",
+                success: function(response) {
```
tidusjar: There is a bug here where if we keep pressing the Get RootFolders it will keep appending to the list :)

PlexRequests.UI/Views/Admin/Sonarr.cshtml
```diff
@@ -80,12 +80,24 @@
             </div>
 
             <div class="form-group">
+                <div>
+                    <button type="submit" id="getRootFolders" class="btn btn-primary-outline">Get RootFolders <div id="getRootFolderSpinner" /></button>
```
tidusjar: Can we include a space in `RootFolders` so it's `Root Folders` ?

PlexRequests.UI/Views/Admin/Sonarr.cshtml
onedr0p: I noticed that lol, it's not an important fix for right now thou

PlexRequests.UI/Views/Admin/Sonarr.cshtml
onedr0p: Yeah that makes sense :)

onedr0p:
> Now SonarrSettings does not contain a definition for the root folders, so they are not being saved into the database.

In teh SonarrSettings class I have 
```
        public bool Enabled { get; set; }
        public string ApiKey { get; set; }
        public string QualityProfile { get; set; }
        public string RootFolder { get; set; }
        public string RootPath { get; set; }
        public bool SeasonFolders { get; set; }
```

The array of paths will also be stored here with the one the user selected? Or should I change it to:
```
        public bool Enabled { get; set; }
        public string ApiKey { get; set; }
        public string QualityProfile { get; set; }
        public string[] RootFolder { get; set; }
        // public string RootPath { get; set; }
        public bool SeasonFolders { get; set; }
```

onedr0p:
Using those new github features I see :P

tidusjar:
Well you need to know the one the user selected right? 

Ignore my other comment, when saving we need to get it out of the cache and set the one the user picked into the RootFolders property. 

On the approve page we can just request the root folder from Sonarr again.

whitecollar17:
@onedr0p is the pull ready for merge?

tidusjar:
It's not finished yet. I think I'll merge it in and then finish it off :)

onedr0p:
Sorry for the late response, please continue where I left off.

Issue #537, opened 2016-09-14
Swedish header isnt in UTF-8


Christley:
#### Mono Version: no clue what that is

this is how the header looks like if you select swedish: https://s9.postimg.org/6khj2o5pr/2016_09_15_01_24_16.png
the rest of the page displays fine. the header is just encoded wrongly.

also, you might want to add this post about the firewall issue in windows to your FAQ, because right now the FAQ about it is superhard to understand. i spent atleast an hour looking around if someone has solved it online before this gem came up: https://www.reddit.com/r/PleX/comments/4dkt6s/plex_requests_net_v16_now_with_headphones_support/d1si2y9

tidusjar:
Ok thanks. I'll fix that issue. 

I'll look at improving the firewall FAQ too! :)

tidusjar:
Fixed in @madesch 's merge!

PR #564, opened 2016-10-04
Set meta charset to be utf-8

madesch:
This fixes issue #537 where the navigation wouldn't have the correct encoding hence not displaying åäö and other characters correctly.

Solved by setting encoding to be utf-8.


tidusjar:
Can you please change the target branch to dev? 

Thanks a lot!

madesch:
Of course - I've changed the branch now :)

tidusjar:
Thank you very much! 👍

PR #568, opened 2016-10-08
Added Paypalme options, no UI yet

ProtoJazz:
I added the options for custom donate button, but the actual implementation of the button its self Im a little unsure on. 

Id like to add a configurable button that shows up next to the existing donate, and have 2
"Donate to PlexRequests" "Donate to *Library Maintainer"

I can see where the existing one is setup in _Navbar, however since thats a partial thats rendered from the layout, Im not sure how to go about getting the data from settings. everything in navbar comes from Context. I think passing a model into layout would be the wrong choice, another option is just an endpoint somewhere that can get fetched and setup with an ajax call. But that also feels too complicated for a simple task. 

All I want/need to do is check the values of the settings, and render a new button if they exist.

### DIFF
--- a/PlexRequests.UI/Modules/DonationLinkModule.cs
+++ b/PlexRequests.UI/Modules/DonationLinkModule.cs
@@ -50,7 +50,7 @@ namespace PlexRequests.UI.Modules
                 Log.Warn("Exception Thrown when attempting to check the custom donation url");
                 Log.Warn(e);
                 JObject o = new JObject();
-                o["url"] = "https://www.paypal.me/PlexRequestsNet";
+                o["url"] = "donationLinkError";
                 o["message"] = "\"" + "Donate to Library Maintainer" + "\"";
                 return Response.AsJson(o);
             }
--- a/PlexRequests.UI/Modules/DonationLinkModule.cs
+++ b/PlexRequests.UI/Modules/DonationLinkModule.cs
@@ -28,31 +28,23 @@ namespace PlexRequests.UI.Modules
 
         private async Task<Response> GetCustomDonationUrl(ISettingsService<PlexRequestSettings> pr)
         {
-            PlexRequestSettings settings = pr.GetSettings();
+            PlexRequestSettings settings = await pr.GetSettingsAsync();
             try
             {
                 if (settings.EnableCustomDonationUrl)
                 {
-                    JObject o = new JObject();
-                    o["url"] = "\""+ settings.CustomDonationUrl + "\"";
-                    o["message"] = "\"" + settings.CustomDonationMessage + "\"";
-                    return Response.AsJson(o);
+                    return Response.AsJson(new { url = settings.CustomDonationUrl, message = settings.CustomDonationMessage });
                 }
                 else
                 {
-                    JObject o = new JObject();
-                    o["url"] = "https://www.paypal.me/PlexRequestsNet";
-                    return Response.AsJson(o);
+                    return Response.AsJson(new { url = settings.CustomDonationUrl, message = settings.CustomDonationMessage });
                 }
             }
             catch (Exception e)
             {
                 Log.Warn("Exception Thrown when attempting to check the custom donation url");
                 Log.Warn(e);
-                JObject o = new JObject();
-                o["url"] = "donationLinkError";
-                o["message"] = "\"" + "Donate to Library Maintainer" + "\"";
-                return Response.AsJson(o);
+                return Response.AsJson(new { url = settings.CustomDonationUrl, message = settings.CustomDonationMessage });
             }
         }
     }
--- a/PlexRequests.UI/Modules/DonationLinkModule.cs
+++ b/PlexRequests.UI/Modules/DonationLinkModule.cs
@@ -1,0 +1,60 @@
+﻿using System;
+using System.Threading.Tasks;
+
+using Nancy;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using NLog;
+
+using PlexRequests.Core;
+using PlexRequests.Core.SettingModels;
+using PlexRequests.Helpers;
+using PlexRequests.UI.Models;
+
+namespace PlexRequests.UI.Modules
+{
+    public class DonationLinkModule : BaseAuthModule
+    {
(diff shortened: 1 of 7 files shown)


tidusjar:
Yeah I see what you mean. 

I think the best approach would be with an Ajax call, we do something similar already to check if there is a new update. 

Regarding the changes, I think it might be better to just let them provide a URL, otherwise we are restricting them to use PayPal.me and I can see there will be another change request to make it customisable. 

what do you think?

ProtoJazz:
Good point on allowing any URL. Ill look at setting it up the same way as the update stuff.

ProtoJazz:
Added in the actual UI for it. Im sure there are some things that could be improved though

For example it pops in, like the update notification does. However during normal browsing and searching that wouldnt happen all that often since the page doesnt always reload the navbar

I thought about adding validation for the url to make sure it was an actual link, but I don't think that's really needed. Its up to the user to make sure they have a valid url.

PlexRequests.UI/Modules/DonationLinkModule.cs
```diff
@@ -0,0 +1,60 @@
...
+            Get["/", true] = async (x, ct) => await GetCustomDonationUrl(pr);
+        }
+
+        private ICacheProvider Cache { get; }
+
+        private static Logger Log = LogManager.GetCurrentClassLogger();
+
+        private async Task<Response> GetCustomDonationUrl(ISettingsService<PlexRequestSettings> pr)
+        {
+            PlexRequestSettings settings = pr.GetSettings();
```
tidusjar: We could use the `async` method here, `GetSettingsAsync`

PlexRequests.UI/Modules/DonationLinkModule.cs
```diff
@@ -0,0 +1,60 @@
...
+            try
+            {
+                if (settings.EnableCustomDonationUrl)
+                {
+                    JObject o = new JObject();
+                    o["url"] = "\""+ settings.CustomDonationUrl + "\"";
+                    o["message"] = "\"" + settings.CustomDonationMessage + "\"";
+                    return Response.AsJson(o);
```
tidusjar: This I think would be a bit easier to read if it was a dynamic object? e.g.

```
return Response.AsJson(new { url = settings.CustomDonationUrl, message = settings.CustomDonationMessage});
```
Since `AsJson` will automatically serialise the object for you,

PlexRequests.UI/Modules/DonationLinkModule.cs
```diff
@@ -0,0 +1,60 @@
...
+                }
+                else
+                {
+                    JObject o = new JObject();
+                    o["url"] = "https://www.paypal.me/PlexRequestsNet";
+                    return Response.AsJson(o);
```
tidusjar: Same as above

PlexRequests.UI/Modules/DonationLinkModule.cs
```diff
@@ -0,0 +1,60 @@
...
+                }
+            }
+            catch (Exception e)
+            {
+                Log.Warn("Exception Thrown when attempting to check the custom donation url");
+                Log.Warn(e);
+                JObject o = new JObject();
+                o["url"] = "https://www.paypal.me/PlexRequestsNet";
+                o["message"] = "\"" + "Donate to Library Maintainer" + "\"";
+                return Response.AsJson(o);
```
tidusjar: same as above

PlexRequests.UI/Views/Shared/Partial/_Navbar.cshtml
```diff
@@ -91,5 +93,22 @@
             </ul>
         </div>
     </div>
+    <script>
+        $.ajax({
+            url: "/customDonation",
```
tidusjar: We will have to make sure the `url` also includes the baseUrl. Take a look at some of the existing Javascript that does this (e.g. `SonarrSettings.cshtml` view)

PlexRequests.UI/Views/Shared/Partial/_Navbar.cshtml
```diff
@@ -38,10 +39,11 @@
                 {
                     <li><a id="donate" href="https://www.paypal.me/PlexRequestsNet" target="_blank"><i class="fa fa-heart" style="color: red"></i> @UI.Layout_Donate</a></li>
                 }
+                <li id="customDonate" style="display: none"><a id="customDonateHref" href="https://www.paypal.me/PlexRequestsNet" target="_blank"><i class="fa fa-heart" style="color: yellow;"></i> <span id="donationText">Donate to Library Maintainer</span></a></li>
```
tidusjar: Can we add the text into the Resources? Similar to what we have done above `@UI.Layout_Donate`

PlexRequests.UI/Modules/DonationLinkModule.cs
ProtoJazz: Fixed

PlexRequests.UI/Modules/DonationLinkModule.cs
ProtoJazz: I tried that at first but it didn't seem to work. I must have made a mistake somewhere because it works now.

PlexRequests.UI/Views/Shared/Partial/_Navbar.cshtml
ProtoJazz: I added it to the resources. Seems like if it doesn't exist in translations it just defaults to English? I that just how it is for now, and at some point someone translates it, or should I just run it through google translate? I'm not sure how localization works for open source stuff, for work we get it done in batches of thousands of strings at a time, but sometimes I'm not sure how much better than google it is.

PlexRequests.UI/Views/Shared/Partial/_Navbar.cshtml
ProtoJazz: Fixed

PlexRequests.UI/Views/Shared/Partial/_Navbar.cshtml
tidusjar: That's fine, usually people edit a spreadsheet that I maintain and I import it

tidusjar:
Awesome work! are these your final changes? If so i'll merge it in :)

ProtoJazz:
I think that's it, unless there's something I've missed.

Issue #586, opened 2016-10-11
{"result":true,"message":null} on landing page setup in 1.9.4


RXWatcher1:
#### Plex Requests.Net Version:
1.9.4

#### Operating System:
Ubuntu 16.04

#### Mono Version:
4.6.1

#### Applicable Logs (from `/logs/` directory or the Admin page):
Not seeing any entries related to this in the logs

#### Reproduction Steps:
using a base URL of /request

1) Log into admin section
2) enable landing page
3) Click Submit

It seems to have saved the changes however the landing page never shows up.

JimboMonkey1234:
Repros for me as well. Step (2) isn't required however, clicking the submit button with no changes has the same result.

Created a pull request: #597 

Haven't tested yet, but I'm pretty sure that's it.

whitecollar17:
Fixed in the release of 1.9.4

RXWatcher1:
You mean fixed in 1.9.5? It's not fixed in the 1.9.4, thats where the issue was found.
Is the 1.9.5 release pending?

tidusjar:
Tonight @RXWatcher1 ;)

whitecollar17:
Yes I'm sorry. I mean  1.9.5.  The release is very soon

RXWatcher1:
This issue remains in 1.9.5.

Mono: 4.6.1
Ubuntu 16.04
Base url is set to "plexrequests"

Full URL: https://my_domain/plexrequests/admin/landingpage

web page output: {"result":true,"message":null}

SuperPotatoMen:
@RXWatcher1 
Can you confirm version 1.9.6 fixed the problem?

RXWatcher1:
I thought I commented but apparently I didn't. I was focused on the --base issue.

 This does fix the {"result":true,"message":null} message however the landing page behavior is erratic. I havent nailed down when it shows up. It seems most of the time it just doesnt show up but sometimes it will show up when logging out of the admin section. I am using the --base build but I experienced the same issue with 1.9.6.

I have it set to show before the login and I have the message enabled with no time limited(defaults in place).

SuperPotatoMen:
@RXWatcher1 
Good! can you open a new issue for that?

PR #597, opened 2016-10-16
Fix boostrapper-datetimepicker imports (#586)

JimboMonkey1234:
When the base url is set, resource file imports need a forward slash at
the beginning. Since it's missing from the datepicker.min.css link,
the Landing Page fails to load correctly causing odd behavior.


JimboMonkey1234:
Haven't set up the build system so I can't test it fully, but I tried it out using Fiddler and it seems to work.

tidusjar:
Thanks a lot!

PR #647, opened 2016-11-02
Update appveyor.yml

tidusjar:


SuperPotatoMen:
Status page don't show there is a update, i miss the patch: true in the appveyor.yml file ?

PR #843, opened 2016-12-27
Fix small typo in ticket overview page

mhann:
Just a small change to put a space between the colon and the admin message to fix:

screen shot 2016-12-27 at 19 26 21

Untested but such a small change I cannot see it breaking anything.


mhann:
Huh, it appears that travis-ci build is broken... My knowledge of c# is not enough to effectively debug. Any ideas?

tidusjar:
It's an issue with the renane of the solution. It's fine. Thanks though!!

mhann:
Ah, makes sense :) np

PR #1094, opened 2017-02-10
Break out Mass Email feature into its own settings tab

dhruvb14:
Also cleaned up some comments and upgraded to the newest font awesome package for more icons :-)


tidusjar:
Thanks!!

PR #1174, opened 2017-02-22
Added permission to see who made requests.

distaula:
I've created a user permission that allows non-admins to see who made the requests.

If the site is configured to allow login without a password, this will expose how to log in as other users, but I don't think that should be a consideration.


distaula:
I made the change to movies and swapped the target from master to dev but it looks like there were a few files changed in the Master branch.  How should that be handled?

tidusjar:
In what way?

distaula:
The Readme and Issue_Template changes became part of the pull-request when I changed targets.  I re-set the branch to the release tag and force-pushed.  The PR should now only contain changes for the feature.

distaula:
@tidusjar I believe that the changes requested are resolved, does it look right?

tidusjar:
Sorry for the delay! 

Looks good, thank you!

PR #1178, opened 2017-02-23
FEATURE: Search movies by actor

smcpeck:
Added a new tab next to "Movies" for "Actors."

Added an `/actor/` API endpoint that maps back to a call on the TMDdAPI Client.

What I don't like is that I have to loop through each movie from an actor's credits and read it individually.  I've limited it to only return 10 movies per search.  But this kinda beats on the API more than I think is ideal.  I don't see a TMDbAPI endpoint that lets me pass in an array of movie IDs. :(

Also, I kinda fumbled through the javascript until I got things working.  I don't know if how I did things is the best approach and that should get some in depth review from you guys.  It's a little weird because it creates a paradigm where one setting `SearchForMovies` results in 2 tabs for the user.  But I think it works decently well as I have it.

Next up would be to do the same thing for TV Shows.

Potential bonus:  I added some `await Task.Yield()` calls in functions that were async, but running synchronously.  I have a theory that this might help some of the odd Thread exceptions crashing Ombi on mono.  I'm running on mono and will see how things look locally on a build of this.


SuperPotatoMen:
First of all, Thank you for your pr!  I will ask Jamie to look at your code.

Ombi.Api/TheMovieDbApi.cs
```diff
@@ -100,5 +105,33 @@ public TmdbMovieDetails GetMovieInformationWithVideos(int tmdbId)
             var movies = await Client.GetMovie(imdbId);
             return movies ?? new Movie();
         }
+
+        public async Task<List<Movie>> SearchActor(string searchTerm)
+        {
+            SearchContainer<SearchPerson> result = await Client.SearchPerson(searchTerm);
+            var person = result?.Results[0] ?? null;
```
tidusjar: Why are we only grabbing the first result here?

Ombi.UI/Modules/SearchModule.cs
```diff
@@ -233,15 +240,10 @@ private Response GetTvPoster(int theTvDbId)
...
-        private async Task<Response> ProcessMovies(MovieSearchType searchType, string searchTerm)
-        {
-            List<MovieResult> apiMovies;
 
-            switch (searchType)
-            {
-                case MovieSearchType.Search:
-                    var movies = await MovieApi.SearchMovie(searchTerm).ConfigureAwait(false);
-                    apiMovies = movies.Select(x =>
+        private List<MovieResult> TransformSearchMovieListToMovieResultList(List<TMDbLib.Objects.Search.SearchMovie> searchMovies)
```
tidusjar: I agree, the whole search module grown organically with not much cleanup.

tidusjar:
Thanks a lot man! The code looks good, but i'll pull down your branch and do some testing first 👍

Ombi.Api/TheMovieDbApi.cs
smcpeck: Sometimes, more than one person could be returned.  They **seem** to return the most popular actor first, so this just grabs that person.  I was looping through the actors at first and hitting each of them, but (obviously) that causes even more hits to the API.  This was basically an easy hack as part of a first pass.

tidusjar:
OK! Did a bit of testing and it looks really good!

Noticed one thing, when we search for a movie or TV show and there are no results we display a small message saying no results. I can see you added the code for that but it doesn't seem to be working.

Second thing, you added the ` SearchForActors` property in the PlexRequestsSettings but it's not on the Settings page (`Admin\Settings.cshtml`)?
Can we add it to the settings page so the users can turn this feature on and off if they wish?

smcpeck:
Yeah - I'll look into those things and update the PR.  Should be out there for you by morning (for you).

Ombi.Api/TheMovieDbApi.cs
```diff
@@ -100,5 +105,48 @@ public TmdbMovieDetails GetMovieInformationWithVideos(int tmdbId)
...
+        {
+            return await SearchActor(searchTerm, null);
+        }
+
+        public async Task<List<Movie>> SearchActor(string searchTerm, Func<int, string, string, Task<bool>> alreadyAvailable)
+        {
+            SearchContainer<SearchPerson> result = await Client.SearchPerson(searchTerm);
+            
+            var people = result?.Results ?? new List<SearchPerson>();
+            var person = (people.Count != 0 ? people[0] : null);
```
smcpeck: This fixes the issue with the lack of a "no results" message.  The old way I was doing this would try getting element 0 of an empty result set and the API call was 500'ing to the client... so yeah.

smcpeck: @tidusjar ^

smcpeck:
New push fixes the issues mentioned.

With the latest merge, I noticed that my stuff won't work for the Emby integration.  I can look into that at some point soon, too.

Ombi.UI/Modules/SearchModule.cs
```diff
@@ -222,6 +229,36 @@ public class SearchModule : BaseAuthModule
...
+            return await TransformMovieResultsToResponse(movies);
+        }
+
+        private async Task<bool> AlreadyAvailable(int id, string title, string year)
+        {
+            await Task.Yield();
+            return IsMovieInCache(id, String.Empty) || PlexChecker.IsMovieAvailable(plexMovies(), title, year);
+        }
+
+        private IEnumerable<PlexContent> plexMovies()
```
smcpeck: My goal here was to cache the results from the Plex repository as it seemed it was getting called a lot.  Maybe that's a necessary evil?

In any event, with my new checkbox for "only show new results," I wanted a subsequent search to filter out items that were recently requested.  That isn't working, so maybe I'm going about it in the wrong way here.

Any thoughts, @tidusjar?

Ombi.Core/SettingModels/PlexRequestSettings.cs
```diff
@@ -41,6 +41,7 @@ public PlexRequestSettings()
         public int Port { get; set; }
         public string BaseUrl { get; set; }
         public bool SearchForMovies { get; set; }
+        public bool SearchForActors { get; set; }
```
smcpeck: Feature enable flag added to UI and respected by the appropriate View.

Ombi.UI/Modules/SearchModule.cs
```diff
@@ -297,22 +323,25 @@ private List<MovieResult> TransformMovieListToMovieResultList(List<TMDbLib.Objec
             return await TransformMovieResultsToResponse(apiMovies);
         }
 
-        private async Task<Response> TransformMovieResultsToResponse(List<MovieResult> movies)
+        private async Task<Dictionary<int, RequestedModel>> requestedMovies()
```
tidusjar: Can we use CamelCasing please?

Ombi.UI/Modules/SearchModule.cs
```diff
@@ -216,6 +219,29 @@ public class SearchModule : BaseAuthModule
...
+            return await TransformMovieResultsToResponse(movies);
+        }
+
+        private async Task<bool> AlreadyAvailable(int id, string title, string year)
+        {
+            await Task.Yield();
+            return IsMovieInCache(id, String.Empty) || Checker.IsMovieAvailable(plexMovies(), title, year);
+        }
+
+        private IEnumerable<PlexContent> plexMovies()
```
tidusjar: Same here with CamelCasing

Ombi.UI/Modules/SearchModule.cs
```diff
@@ -216,6 +219,29 @@ public class SearchModule : BaseAuthModule
...
+        {
+            if(_plexMovies == null)
```
tidusjar: This is a good idea to use the backing field here. Saves the db calls.

Ombi.Api/TheMovieDbApi.cs
```diff
@@ -100,5 +105,48 @@ public TmdbMovieDetails GetMovieInformationWithVideos(int tmdbId)
...
+                            break;
+                        if (alreadyAvailable == null || !(await alreadyAvailable(credit.Id, credit.Title, credit.ReleaseDate.Value.Year.ToString())))
+                        {
+                            movies.Add(await GetMovie(credit.Id));
+                            counter++;
+                        }
+                        else
+                        {
+                            //if we didn't add the movie, delay longer since we know we'll be making additional API calls
+                            await Task.Delay(75);
```
tidusjar: We are delaying the task to potentially cope with the API call load?
Since if it falls here it will be delayed by 75ms and then an additional 50ms? Is this expected?

Ombi.UI/Modules/SearchModule.cs
tidusjar: Why filter them out? We don't really do that anywhere else in the application, we just show "Requested" when a search result has already been requested

Ombi.UI/Views/Search/Index.cshtml
```diff
@@ -72,8 +79,28 @@
...
+                        <!-- Actors tab -->
+                <div role="tabpanel" class="tab-pane" id="ActorsTab">
+                    <div class="input-group">
+                        <input id="actorSearchContent" type="text" class="form-control form-control-custom form-control-search form-control-withbuttons">
+                        <div class="input-group-addon">
+                            <i id="actorSearchButton" class="fa fa-search"></i>
+                        </div>
+                    </div>
+                    <div class="checkbox">
+                        <input type="checkbox" class="newMoviesOnly" id="actorsSearchNew" name="actorsSearchNew"><label for="actorsSearchNew">@UI.Search_NewOnly</label>
```
tidusjar: Like I mentioned above, not sure if this is even required.

tidusjar:
Looks good, my only concern is like you mentioned, we have removed the Emby integration from the search pages now.

Ombi.UI/Modules/SearchModule.cs
smcpeck: Can definitely remove this, I just thought it would be neat since I feel the use case is a bit different than searching straight for a movie.  Like, when searching for a movie, I have a title in mind and I want to find it.

When searching for a movie, I'm probably more thinking, "What movies is So-and-so in that I don't have yet?"

If we only ever show the first 10, it loses some longer term value.  I suppose the **real** problem is that we are only showing 10 and don't offer a way to page through all results.  So this was an attempt to get more than the first 10.

If you think it's too different, I can stash this stuff aside for later consideration.

Ombi.Api/TheMovieDbApi.cs
smcpeck: Yeah.  It's a bit arbitrary.  Basically, if we are making 10 calls, we sleep 50ms.  If we make more than 10, we sleep 125ms for every extra call after that.

This definitely slows the search a bit and I'd love to lower the delays.  Maybe I'll rip that extra 75ms out, there's not really any rhyme or reason behind it -- just didn't wanna beat the hell out of the API and hit the rate limit.

1 - What is the API rate limit?
2 - Is the same API key being used across everyone's installation of Ombi?

Ombi.UI/Modules/SearchModule.cs
```diff
@@ -216,6 +219,29 @@ public class SearchModule : BaseAuthModule
             return await TransformMovieResultsToResponse(movies);
         }
 
+        private async Task<Response> SearchActor(string searchTerm, bool filterExisting)
```
tidusjar: `filterExisting` is not being used anywhere?

Ombi.UI/Modules/SearchModule.cs
tidusjar: I think we should stick to the way the search currently works. It's a good idea, but I think it might bring on some confusion.

Ombi.UI/Modules/SearchModule.cs
```diff
@@ -222,6 +229,36 @@ public class SearchModule : BaseAuthModule
...
+
+        private async Task<bool> AlreadyAvailable(int id, string title, string year)
+        {
+            await Task.Yield();
+            return IsMovieInCache(id, String.Empty) || PlexChecker.IsMovieAvailable(plexMovies(), title, year);
+        }
+
+        private IEnumerable<PlexContent> plexMovies()
+        {   long now = DateTime.Now.Ticks;
+            if(_plexMovies == null || (now - _plexMovieCacheTime) > 10000)
```
tidusjar: What's this for `(now - _plexMovieCacheTime) > 10000` ?

Ombi.Api/TheMovieDbApi.cs
tidusjar: API key is per IP address
Rate is 40 requests every 10 seconds.

I think it's a good idea to have the delay, I just wanted to make sure there was a reason for it rather than let's just delay it 😄

Ombi.UI/Modules/SearchModule.cs
tidusjar: Actually the checkbox is there and does make it obvious... Ok leave it 😆

Ombi.UI/Modules/SearchModule.cs
smcpeck: It'll update the `_plexMovies` cache every 10 seconds.

Ombi.Api/TheMovieDbApi.cs
smcpeck: Given the API key/rate information, I'm going to rip out the extra 75ms.  It would actually be neat to build in some throttling at the TMDbAPI level. How active is that project?

smcpeck:
New push removes the extra delay.  I **think** we're ready.  Is there a place to add some documentation for the user to let them know some details (like: `Don't go crazy or you'll hit the API rate limit.` and to kinda explain the checkbox)?

My next PR will be to clean up `SearchModule` a bit.

tidusjar:
Can we add the code back in to check if the content is on Emby (don't worry about testing it, i can do that)?

smcpeck:
Yup.  I'll work on that now and clean up a couple other minor things I've noticed.

Ombi.UI/Modules/SearchModule.cs
```diff
@@ -243,11 +247,15 @@ public class SearchModule : BaseAuthModule
...
         private async Task<bool> AlreadyAvailable(int id, string title, string year)
         {
-            await Task.Yield();
-            return IsMovieInCache(id, String.Empty) || PlexChecker.IsMovieAvailable(plexMovies(), title, year);
+            var plexSettings = await PlexService.GetSettingsAsync();
+            var embySettings = await EmbySettings.GetSettingsAsync();
+
+            return IsMovieInCache(id, String.Empty) ||
+                (plexSettings.Enable && PlexChecker.IsMovieAvailable(PlexMovies(), title, year)) ||
+                (embySettings.Enable && EmbyChecker.IsMovieAvailable(EmbyMovies(), title, year, String.Empty));
```
smcpeck: This is the main chunk intended to filter based on Emby content.

smcpeck:
I think I'm ready to call things done.  Let me know if you catch anything else.

SuperPotatoMen:
@smcpeck you closed it to soon, otherway we can't merge it in to the dev branch 😆

smcpeck:
Oops.  I'll take a look a little bit later.  Lunchtime for now. 🍔

smcpeck:
Hmmm.  What am I missing?
image

smcpeck:
Is there a way to force the TravisCI job to rerun?  The mono build dependencies were spitting back a 503 at the time the job ran.

tidusjar:
Restarted that build. Sometimes that happens

tidusjar:
So all good to merge in? 😄

smcpeck:
Yup.  Merge away!

tidusjar:
Excellent job dude!

smcpeck:
Thanks.  This stands as my first opensource contribution. 🎉 
I look forward to adding some extra functionality and fixing a few issues.
Cheers!

ketsapiwiq:
Hi!
I was looking at adding a "search movies by director" feature, and I can't get where the code differs! It follows TMDB API for a `person` even though in your code it's referred as an `actor` all along...
If anybody has a clue.
Thanks!

smcpeck:
Hey, @ketsapiwiq -- sorry so late here. Please note that this PR/feature is over 5 years old and I think Ombi has had a full rewrite since then...

That said, take a look here: https://github.com/Ombi-app/Ombi/pull/1178/files#diff-eae5bd394beea017ab3bb11264fb23006b3d99faf4e5e9fabbb8204fc119adb8R131

Movies are first grabbed where there was a match for "cast" and then things are filled in on the next line with any "crew" matches (so this would get directors, yes). After that, there's a little filtering loop so we get the first 10 movies which are not already available and then display them to the user.

Hope this helps.

ketsapiwiq:
Thanks a lot, I found your code is still there: https://github.com/Ombi-app/Ombi/blob/develop/src/Ombi.Core/Engine/MovieSearchEngine.cs#L87

You made me realize you can already search for a director's films if you go to the `/discover/actor/<id>` route!

Then I guess the only additions needed would be on the front-end side, in the view corresponding to the route `/details/movie/<id>`, showing the director's name with a link to `/discover/actor/<id>`! :)

screamjojo:
Hi,
I'm also interested, in first I would like to see the "director" on each movie details, and then having the possibility to click on it like an actor ...
Do you have a solution ?
Thanks in advance.

ketsapiwiq:
I don't have a solution, but what we need to ask is to show the "Crew" info on a movie details page, and not only the "Cast".
The crew will include the director(s) and have a clickable link.

ketsapiwiq:
Actually @screamjojo I created PR #4722 for it, if you're nerdy enough you can use this version :)

PR #1251, opened 2017-03-14
Update German Translation

Lauch:
I updated the German translation with correct grammar and slightly more correct words.


tidusjar:
Excellent! thanks!

Can you target the dev branch rather than master please?

Issue #1286, opened 2017-03-26
If email settings incorrect test message still says successful


dhruvb14:
#### Ombi Version:

 Commit 3be69cb Dev Branch

#### Update Branch:

development

#### Media Sever:

N/A

#### Media Server Version:

N/A

#### Operating System:

Docker/Mono

#### Mono Version (only if your not on windows)

4.0.30319.42000

#### Applicable Logs (from `/logs/` directory or the Admin page):

N/A

(Logs go here. Don't remove the ``` tags for showing your logs correctly. Please make sure you remove any personal information from the logs)

N/A

#### Problem Description:

If you put in wrong hostname or wrong settings for email async task and it hits exception, it still sends a success message to user's UI.

#### Reproduction Steps:

Enable Email notifications for smtp.google.com (its actually smtp.gmail.com) and fill out all other information. Click test and notice it says success however email never sends. Line 212 in EmailMessageNotification.cs causes exception since smtp.google.com is not a valid smtp endpoint. Code hits exception on 229 but task is "Successful" so nancy responds with success message.

PR #1287, opened 2017-03-26
Quick fix for email false positive message. ISSUE: #1286

dhruvb14:
Also VS2017 without Update 1 complains about every cshtml file until updated. Just a FYI incase you updated and see the issue.


tidusjar:
Ah I see what you mean now.

The `Send` property is true when there are movie items or tv items. So if it's false and there's a test email then it will be empty. But I guess we should still send it.

For test emails that have no content perhaps we should `Take(10)` just to get content in for movies and shows?

dhruvb14:
I guess we could take(10) but then if there was some issue we would not know since the test email and normal email would have different code paths.

tidusjar:
Very true.

Is this ready to be merged in? 😄

dhruvb14:
Yes merge ready if you approve :-)

PR #1358, opened 2017-04-07
Remove unneeded bundle config

MattJeanes:
With the new bundling config stuff this is no longer needed


tidusjar:
@MattJeanes is amazing. ❤️

PR #1420, opened 2017-06-04
lossless compression of images saved 90KB

Fish2:
Save Bandwidth and speed up website load time but lossless completion of all images.

doing this saved 90KB docent sound like much but why not


tidusjar:
Thanks!

Issue #1410, opened 2017-05-07
Wrong translate in language: Swedish


hakannordell:
#### Applicable Logs (from `/logs/` directory or the Admin page):

```

(Logs go here. Don't remove the ``` tags for showing your logs correctly. Please make sure you remove any personal information from the logs)

```

#### Problem Description:

When you search, you get a list of Movies. On the right hand you have a button "+Request" in English. Thats fine, but if you choose the Swedish language, it says "+Problem".  That means "+Truble" :-)
It should say "+Önska".

#### Reproduction Steps:

Please include any steps to reproduce the issue, this the request that is causing the problem etc.

PR #1450, opened 2017-06-28
Changed two words that was wrong

rabbadab:
One misspelled word in the main menu. Önsningar -> Önskningar
And changed Problem to Önska according to the issue: #1410


tidusjar:
Thanks!

Issue #1457, opened 2017-07-05
Requests


tidusjar:
- Rework the TV Requests page
    - This still need a lot of work, mainly around how we will visually display how to approve "child" requests, I was thinking about when we display the TV requests there should be a single button "Manage" that then takes to a new page listing all child requests where we can then, approve, decline, delete etc.

tidusjar:
Anyone have any idea (UI design wise) how this could look?

dhruvb14:
Working on this now, have it partially working but can't get tvshowrequestbuilder.cs to stop throwing exception. Once fixed all that is left is CSS, however with this new bug I can't test CSS till C# code is fixed

Video: https://1drv.ms/v/s!AutEqhj08rK0hPZdfbCuyFTiCPaigA
If you want to help, here is the code github.com/dhruvb14/ombi

Once fixed ill create a PR

dhruvb14:
Progress So far: https://1drv.ms/v/s!AutEqhj08rK0hPZhpM_dxR630bdKuA
#1488

tidusjar:
Looks great!

tidusjar:
@dhruvb14 is the `tvrequests-manage` component needed anymore?

dhruvb14:
@tidusjar I want to leave it for now since I plan on embedding that component as a partial view into the main screen to clean up code. Just had not gotten around to it in initial PR. Code in that file and the new view I created is almost 1:1 so it would be great code-reuse/maintainability feature to leave.

tidusjar:
@dhruvb14 Is there any point me hooking up all the buttons and sorting that out yet?

dhruvb14:
If you do, make sure you pass around objects in the functions as it makes its 2x easier to move functions around. (ng-click="addChildren(children)" instead of ng-click="addChildren()") Don't use any stored variable on the component from inside any functions you call as the new view has multiple children visible. I plan on trying to clean up code this weekend also so maybe I can make use of that old manage view by embedding it into the new UI so it's easier to maintain.

tidusjar:
Yeah that makes sense, I'm planning to work on the scheduled jobs, so things appear as available.

dhruvb14:
#1496

Issue #865, opened 2016-12-30
Port to .Net core


tidusjar:
Let's all face it, anyone running on anything but windows usually runs into issues at some point.

Let's port this over to .Net core and remove the dependency from Mono.

vertigo235:
So is it not planned yet or in progress?  I'm confused

tidusjar:
It's in progress but a backburner, only working on it when I have time.

This will take a while to port everything over, since the frameworks are different, Core API is different from the Full Framework too and there are many dependencies that we use are not supported on .Net Core yet.

yeyeoke:
Wow, truly great news, I'm running Ombi on Ubuntu Server 16.10 and I've been experiencing that Ombi becomes unresponsive and only comes online again after I restart it, I figured it was due to mono.

tidusjar:
Yeah, we have few issues from the limitations of mono. It's an excellent framework but it does have its caviats! 

This has been put on hold for the time being as I am waiting for .Net core to mature a bit as there are still problems with the tooling and it's still very much influx.

RyanAStrouse:
Excellent news! Thanks!

SuperPotatoMen:
@tidusjar just started with tis. it will take a while but i think the linux users will be happy :+1:

tidusjar:
Sonarr Integration is taking a while

Few reasons:

- There are many different scenarios to test e.g. Single Episode request, Multiple episode requests in different seasons, Full season request, whole series request etc.

- To cater for the above scenarios I need to code this right 😄 

- I want to get it right the first time around!

roly445:
Do you need any help with this?

tidusjar:
I could always use help!

roly445:
Do you have a task list.  Should I email you?

bulbajackel:
+1 I would like to help with some of the angular 4 work that you have.

tidusjar:
@roly445 I'll create a task list, it's all in my head currently!

@bulbajackel  yeah man that's cool!

tidusjar:
I've created a bunch of high level tasks that is required! If you have any questions post on them :)

RaymondSchnyder:
How far is this progressed? I ask because ombi crashes every 2 minutes on mono 5 xD

nemchik:
There were 9 commits in the past 10 days. Not sure if they were game changers, but there is some kind of progress.

I do not have issues with mono 5 other than Ombi being slow as hell. You might have some other issue.

[...]

PR #1488, opened 2017-08-19
Update Request Management UI

dhruvb14:
Video: https://1drv.ms/v/s!AutEqhj08rK0hPZhpM_dxR630bdKuA

I really hated using arrows for the UI and like how clean the "View" button for expanding the child requests was so I had to write a little bit of TS code to manipulate the UI. Asides from that it looks and feels very similar but keeps people on the same page. Also none of the functions as I could see for the approval were actually hooked up so I just copied over the functions from the old approval flow screen. They will have duplicates which can be deleted once we hook up the functions.

#1457 #865


dhruvb14:
Not sure why build failed, looks like issue with build script. Errors about Cake...

dhruvb14:
Yeah it should be. something happened when i merged down the code from upstream, i missed it. Let me update

PR #1496, opened 2017-08-26
Break out Request Children into their own component make Search UI similar to Requests UI

dhruvb14:
Title says it all, no functionality changes just code...


dhruvb14:
Refactored Search to match Requests UI so you can choose episodes inline without leaving the search page, Temporarily removed infinite scroll on requests management screen since it has a bug which I was unable to resolve immediately...

src/Ombi/ClientApp/app/requests/tvrequests.component.ts
```diff
@@ -227,13 +229,11 @@ export class TvRequestsComponent implements OnInit, OnDestroy {
         }
     }
     private removeChildRequestFromUi(key: IChildRequests) {
-        this.tvRequests.forEach((val) => {
```
tidusjar: Do we no longer need this?

dhruvb14: Checked in new update removing unused functions and updated type interface on a missed typed

tidusjar:
Let me know when this is ready to merge on! :)

dhruvb14:
It’s ready

tidusjar:
Excellent!!

PR #1508, opened 2017-09-17
Latest practices... also probably broke some styles - sorry

### LINKED ISSUE #1507: Styles different in the CI Build
No Idea why, cannot produce locally...

Some reason there are missing styles when building via AppVeyor... Using the same commands and Cake script locally the styles are fine...

Might need the master that is @MattJeanes assistance on this one.

MattJeanes:
Fixes #1507 to some degree - is now also broken in localhost lol

- Fixed C# warnings
- Fixed removeRequestFromUi (still might not work but definitely did not before, TS 2.5 showed that the call was just completely wrong before)
- Upgrade to TS 2.5
- More efficient local builds + will fail if webpack compilation fails
- Moved styles to ClientApp/styles for HMR support
- Moved loading css into wwwroot to ensure it loads as fast as possible (sidenote: may need some more styling in there for background color etc now as it loads before everything else)
- Upgraded all packages (except zone.js latest version is borked)
- Using bootstrap from npm now (maybe the cause of issues below?)

Some styles appear to have changed from current appveyor downloadable version to local version, not sure where the styles are so I was hoping you'd know

Current build:
image

This one:
image

Uhh, you're welcome?

Not expecting a merge yet, but maybe you know what might be causing these issues so we can fix them?


MattJeanes:
AppVeyor doesn't seem to like building pull requests.. maybe see if you can fix the version it tries to use  - 1.0.0-pull-1508 should be valid

```
C:\Program Files\dotnet\sdk\2.0.0\NuGet.targets(102,5): error : '1.0.0-pull/1508/merge' is not a valid version string.
```

MattJeanes:
Tried a couple things to fix build but its persistent on using that weird version, talk to ya tomorrow about it

PR #1514, opened 2017-09-23
Optimized images, Update old compressed image with a new lossless one.

camjac251:
I've separated the commits to hopefully easily show each step but I basically optimized all images found old and new with FileOptimizer and without losing any of the quality keeping them 1:1. Metadata was stripped if any was found but I don't think it's really necessary with web images.

I've also replaced the emby-logo-dark.jpg which had bad aliasing and some color banding with a fresh matching lossless one that looks nicer and I updated all of the file paths that I could find to reflect the new image.


tidusjar:
Looks good! :)

PR #1529, opened 2017-10-04
Auto-scale large images down to container size

Unimatrix0:
Automatically scales images larger than the containing div down to the size of the div.


tidusjar:
Thanks again!

PR #1559, opened 2017-10-10
Change date format in the TV Episode view 

jpeters:
Change date format in the TV Episode view to be more consistent with the rest of the date formats in the UI.


tidusjar:
Hmmm, I did think about this.

I personally dislike US dates and prefer `dd/mm/yyyy` or `dd mmm yyyy`. 

I thought about changing it depending on their browsers language e.g. `en-US` would be US date format where `en-GB` would be the *better* format 😉

jpeters:
That makes sense. It can be slightly jarring to see a non region-specific format. That was actually part of the motivation for the PR. The format in place was not my native format, but I liked the format on the rest of the app, so I thought I would attempt to get them in sync. Go ahead and decline the PR and we can work on a better solution.

PR #1676, opened 2017-11-03
Lossless Compression of images saves 83 KB

Fish2:
Lossless Compression of images saves 83 KB optimise page load times


tidusjar:
THANKS!

PR #1778, opened 2017-12-05
fix startup config error

bazhip:
https://github.com/ReactiveX/rxjs/issues/3151#issuecomment-348730503


tidusjar:
Thanks! was just about to do this!

PR #1817, opened 2017-12-28
 Fix duplicated "Requests" element ID on new Issues link

shoghicp:
Issues link currently has the same element ID as the Requests link. This makes styling hard, and it doesn't seem intended anyway.


tidusjar:
Thanks!

PR #1820, opened 2017-12-29
Fix non-admin issues dropdown

rob1998:
Let non-admin users report an issue on a requested item.
I just moved <div class="dropdown" *ngIf="issueCategories && issuesEnabled"> outside of <div *ngIf="isAdmin">


tidusjar:
Thanks!

PR #1838, opened 2018-01-02
Enable Multi MIME and add alt tags to images

louis-lau:
As promised. It's literally the first C# I've coded and I have no idea what I'm doing :').
It works, but it might not be best practice. So feel free to make changes as necessary.

Just look at that sweet 10/10
http://www.mail-tester.com/web-47ta3

### DIFF
--- a/src/Ombi.Notifications.Templates/Templates/BasicTemplate.html
+++ b/src/Ombi.Notifications.Templates/Templates/BasicTemplate.html
@@ -155,7 +155,7 @@
                                 </tr>
                                 <tr>
                                     <td align="center">
-                                        <img src="{@IMGSRC}" alt="Movie poster" width="400px" text-align="center" />
+                                        <img src="{@IMGSRC}" alt="Poster" width="400px" text-align="center" />
                                     </td>
                                 </tr>
                             </table>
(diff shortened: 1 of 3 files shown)


src/Ombi.Notifications.Templates/Templates/BasicTemplate.html
```diff
@@ -155,7 +155,7 @@
                                 </tr>
                                 <tr>
                                     <td align="center">
-                                        <img src="{@IMGSRC}" width="400px" text-align="center" />
+                                        <img src="{@IMGSRC}" alt="Movie poster" width="400px" text-align="center" />
```
tidusjar: This could also be a TV Show poster, can we just change it to `poster`?

src/Ombi.Notifications/Agents/EmailNotification.cs
```diff
@@ -71,14 +71,24 @@ protected override bool ValidateConfiguration(EmailNotificationSettings settings
             return message;
         }
 
+        private async Task<string> LoadMessage(NotificationType type, NotificationOptions model, EmailNotificationSettings settings)
```
tidusjar: Can we change the method name to something like `LoadPlainTextMessage`? 

It could be a bit confusing since there is another `LoadMessage` method on the `base` with different parameters that `LoadTemplate` ends up calling

louis-lau:
Done 👍

PR #1842, opened 2018-01-02
Update Swedish translation

lisinge:
The Swedish translation is alright but the wording is a bit off.
Words like "Begäran" is more like "Demand" than "Request" so i replaced it with "Efterfrågan" which is a lot better.
I also updated the new strings that were not yet translated.


tidusjar:
Please view this:  https://github.com/tidusjar/Ombi/wiki/Translations

lisinge:
Oh cool

PR #1846, opened 2018-01-03
Fix user mentions in Slack notifications

Alasano:
Adding link_names parameter to ensure that incoming webhooks get properly parsed for user mentions. Without this user mentions remain as plain text.


tidusjar:
Nice! Thanks!

PR #1852, opened 2018-01-05
Better login backgrounds!

louis-lau:
Image now centered so it looks better in general, mostly on small devices though. img changed to div to avoid 1px gray border in chrome and safari.


louis-lau:
Forgot to mention, tested in Chrome, Firefox, Edge and Chrome on android. Might need to be tested on ios safari.

tidusjar:
Looks good!!

PR #1872, opened 2018-01-09
Update README.md

xnaas:
The link for proxy examples changed. :)


tidusjar:
Oops!

Thanks for that :)

xnaas:
It's all good. I really wish GitHub used shortlinks or something so it didn't even need to be worried about.

PR #1882, opened 2018-01-12
Make user management table responsive

louis-lau:
Now I'm not sure where you want the css for this, I just put it at the bottom of base.scss. Tell me if you want it somewhere else 👍 

https://vgy.me/Wec0UI.gif

### DIFF
--- a/src/Ombi/ClientApp/styles/base.scss
+++ b/src/Ombi/ClientApp/styles/base.scss
@@ -865,6 +865,7 @@ a > h4:hover {
     .table-usermanagement {
         /* Force table to not be like tables anymore */
         display: block;
+
         thead, tbody, th, td, tr {
             display: block;
         }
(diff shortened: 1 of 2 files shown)


src/Ombi/ClientApp/styles/base.scss
```diff
@@ -858,4 +858,70 @@ a > h4:hover {
...
+            top: 6px;
+            left: 6px;
+            width: 45%;
+            padding-right: 10px;
+            white-space: nowrap;
+        }
+    }
+    /* Label the data */
+    .td-checkbox:before {
+        content: "Select:";
```
tidusjar: We shouldn't be doing this in the CSS for localization reasons, same with the below

tidusjar:
Nice job! But please see comments 😄

louis-lau:
Better? Better.

src/Ombi/ClientApp/styles/base.scss
```diff
@@ -895,33 +896,9 @@ a > h4:hover {
...
-
-    .td-last-logged-in:before {
-        content: "Last Logged In:";
-    }
-
-    .td-user-type:before {
-        content: "User Type:";
+        /* Label the data */
+        .td-labelled:before {
+            content: attr(data-label)
```
tidusjar: I didn't know you could do that!

tidusjar:
Thanks!

src/Ombi/ClientApp/styles/base.scss
louis-lau: Me neither!

PR #1987, opened 2018-02-17
Fixing incorrect filter translation targets

jonocairns:
image

Current schema:

```
"Requests": {
    ...
    "Filter":"Filter",
  },
  "Filter":{
    ...
    "ClearFilter":"Clear Filter",
  }
```


tidusjar:
Thanks!

Can this be done in the other translation files too?

jonocairns:
@tidusjar It's just the reference to the translation target so it should fix all translations for `Requests.ClearFilter` and `Filter.Filter` (I think!)

tidusjar:
Sorry I meant the other translation files. Crowdin will just recognise them as new translations and we will loose the translated values.

jonocairns:
Sorry, I'm still getting familiar with the setup. I don't think I follow :(

tidusjar:
That's fine!

So in the folder where the translations are, there are other ones. E.g. de.json we need to make the same changes in those files

jonocairns:
Ah, these are references to already existing translations. I've double checked the json files and they're all there.

tidusjar:
@jonocairns I know they are all there, but since you modified the "master" which is the `en` file then you also need to make the same changes in the other translations files.

jonocairns:
I haven't edited the en.json file in this PR

tidusjar:
Sorry i'm following. Can you please change the target branch to DotNetCore

PR #2014, opened 2018-02-27
Css fixes

louis-lau:
So, we were apparently importing base.scss twice. One in main.ts and once in plex.scss. I knew I was seeing the same styles twice!

Also @gold007eye this aligns the table headers to the left.


tidusjar:
Oh nice!

PR #2052, opened 2018-03-09
Added the ability to hide requests that have not been made by that user

tidusjar:


Masterditto:
Hi tidus

Where can I find this function inside OMBI? :)

Masterditto:
Yeah, I should have mentioned I have been in all the settings and menu's that I could think of.

I'am running v3.0.3000 - Maybe u can guide me a bit further? >.< Cant seem to find it...

c010rb1indusa:
Just though you should know that even if you enable this setting, it still reads "Below you can see yours and all other requests, as well as their download and approval status." in the requests menu. That should be removed or altered to "Below you can see all your requests, as well as their download and approval status". Thanks a bunch for all your work.

PR #2091, opened 2018-03-23
Goldenpipes discord user

goldenpipes:
Add a discord User ID to the user management page.


tidusjar:
Thanks for the PR!

But I don't think i'm going to merge this in, regarding these sort of things e.g. Discord Id, Slack Id, Pushbullet device Id. I think it would be better to store them in a seperate table and have a many to one relationship with the user rather than extending the User table.

This will allow us to easily add new "additional ids" for the user as it would be a new record in the database.

goldenpipes:
for sure. I'm just trying to monkey around and see whats what.  I'm pretty sure i missed some things anyway :)

PR #2092, opened 2018-03-23
Dynamic Background Animation

anojht:
Implemented a background fade in/fade out on a ten second interval to cycle backgrounds on landing and login pages.


tidusjar:
The build is failing on some of the linting tasks. 

ClientApp/app/settings/settings.module.ts[143, 32]: file should end with a newline

anojht:
@tidusjar Yes I think this is because the file in the develop branch does not end with a newline. Just double checked the branch from my fork and it's not causing the linting error.

tidusjar:
My bad! i'll take a look

src/Ombi/ClientApp/app/landingpage/landingpage.component.ts
```diff
@@ -41,4 +49,22 @@ export class LandingPageComponent implements OnInit {
             this.mediaServerStatus = x;
         });
     }
+
+    public ngOnDestroy() {
+        clearInterval(this.timer);
+    }
+
+    public cycleBackground() {
+        setTimeout(() => {
```
tidusjar: So the `this.timer` calls `this.cycleBackground` every 10 seconds, and then we have another two timers in here?

Once that clears the background and another that sets it. Why do we have this?

tidusjar (changes requested):
See comments

src/Ombi/ClientApp/app/landingpage/landingpage.component.ts
anojht: Previously I was running into sync issues therefore I added the timeout timers. With the new improvements I made they are redundant, will quickly push a new commit with the timers removed.

anojht (commented):
Addressed the review request and made necessary changes

tidusjar (approved):

tidusjar:
Nice!

Issue #1907, opened 2018-01-17
[LC] Issues Details Page needs some artwork


leram84:
Can you throw the bg and poster in that page? Don't care where or how u style it cause im gonna change it anyway, but i need the content there. If you dont feel like setting it up, feel free to `display: none` on the base css.

anojht:
@leram84 BG and Poster are ready. Please style the page as its currently looking standard. 
@tidusjar Just realized this, remember how providerid field was not being set properly prior to my commit so the issues created prior do not have a value for providerid so they will not load the images. Issues created after the new update will work just fine since I fixed the providerid field being set properly on issue creation.

leram84:
Awesome @anojht !! Think you might wanna take a stab at some of the other [LC] issues? Its not really practical for me to release anything till they are all done, since each change to the ui generally means i have to go back and re-write most of this house of cards lol. But great job either way! Much appreciated :)

anojht:
@leram84 I will see if I can knock out the other LC issues over the next few days, if there are any specific things you want please update the issues as it will be easier for me to implement those as I work on them!

leram84:
oh man... caaaaannnn doooo lol. Nothing will change, but im sure some new things have come up since there have been a lot of additions since i made those issues. You on discord? Feel free to jump in the layercake room in ombi or organizr's server, or just pm me if u want/need anything, but ill go through and make sure there isn't anything else that needs to be added to the existing issues.

anojht:
I am in the Ombi discord my discord name is MrGoldenPhoenix. Sounds good!

PR #2093, opened 2018-03-23
Fix for issue #1907

anojht:
Implemented bg and poster on issue details page as per issue #1907


src/Ombi/ClientApp/app/issues/issueDetails.component.ts
```diff
@@ -85,4 +91,24 @@ export class IssueDetailsComponent implements OnInit {
...
+                    ("url(" + "https://image.tmdb.org/t/p/w1280" + x.backdropPath + ")");
+                this.posterPath = "https://image.tmdb.org/t/p/w300/" + x.posterPath;
+            });
+            
+        } else {
+            this.imageService.getTvBanner(Number(issue.providerId)).subscribe(x => {
+                this.backgroundPath = this.sanitizer.bypassSecurityTrustStyle
+                    ("url(" + x + ")");
+            });
+            this.searchService.getShowInformationTreeNode(Number(issue.providerId)).subscribe(x => {
```
tidusjar: We probably shouldn't be doing this just to get the banner.
This will pull back ALL the information about the request, and we need to keep in mind that there might not be a request associated with the issue.

We can raise issues on content that do not have requests.

tidusjar (changes requested):

src/Ombi/ClientApp/app/issues/issueDetails.component.ts
anojht: @tidusjar Good point, I did not think of that. Would writing a few API calls that just returns image paths be the best solution here instead? For example, I wrote API calls inside image.service.ts but the FanArt banner and poster images were inconsistent with the request or search images. So maybe just writing new API calls for TheMovieDBApi.cs and TvMazeApi.cs to return just the images would work?

tidusjar: Yes, we have the provider id so we can just use the Image service (check the Image Controller to see what we currently support)

If you are writing a new api then I suggest it should be on the ImageController.

anojht (commented):
Addressed review and wrote new calls to just get images

tidusjar (approved):

tidusjar:
Looks good. I can't see anything else here.

Have you done the appropriate dev testing to make sure it's functioning correctly?

tidusjar:
Thanks a lot!

PR #2100, opened 2018-03-24
Created a fancy embed for the "new request" webhook message.

roanvonb:
DiscordWebhookBody .cs - Added variables and classes to represent the discord embed.
DiscordNotification.cs - Fetches details from MovieRequest and TvRequest to fill out the above.

I am a noob so please check if I did anything stupid. More to come once I know I'm not doing something horribly wrong.

### DIFF
--- a/src/Ombi.Notifications/Agents/DiscordNotification.cs
+++ b/src/Ombi.Notifications/Agents/DiscordNotification.cs
@@ -33,7 +33,7 @@ namespace Ombi.Notifications.Agents
 
         // if true mentionAlias will post the alias to discord and trigger an @mention if set up as <@id>.  
         // It will also use the username instead of alias when talking about a user. e.g. "Requested by username on 16 March"
-        public bool mentionAlias { get; set; } = true;
+        public bool MentionAlias { get; set; } = true;
 
         public override string NotificationName => "DiscordNotification";
 
--- a/src/Ombi.Notifications/Agents/DiscordNotification.cs
+++ b/src/Ombi.Notifications/Agents/DiscordNotification.cs
@@ -94,24 +94,19 @@ namespace Ombi.Notifications.Agents
             if (Customization.ApplicationUrl.HasValue())
                 authorUrl = $"{Customization.ApplicationUrl}requests";
 
-            ShowCompactEmbed.TryGetValue(NotificationType.NewRequest, out var compactEmbed);
-            MentionAlias.TryGetValue(NotificationType.NewRequest, out var mentionUser);
-
             DiscordAuthor author = new DiscordAuthor
             {
-                name = "New Request!",
-                url = authorUrl,
+                name = parsed.Subject,
                 icon_url = "https://i.imgur.com/EPuxVav.png"
             };
-
             DiscordEmbeds embed = null;
             if (model.RequestType == RequestType.Movie)
             {
-                embed = createDiscordEmbed(author, MovieRequest, parsed.Image, compactEmbed, mentionUser);
+                embed = CreateDiscordEmbed(author, MovieRequest, NotificationType.NewRequest, parsed.Image);
             }
             else if (model.RequestType == RequestType.TvShow)
             {
-                embed = createDiscordEmbed(author, TvRequest, parsed.Image, compactEmbed, mentionUser);
+                embed = CreateDiscordEmbed(author, TvRequest, NotificationType.NewRequest, parsed.Image);
             }
             await Send(notification, settings, embed);
         }
@@ -187,7 +182,7 @@ namespace Ombi.Notifications.Agents
                 Message = message
             };
             notification.Other.Add("image", image);
-            await Send(notification, settings);
+            await Send(notification, settings);           
         }
 
         protected override async Task RequestDeclined(NotificationOptions model, DiscordNotificationSettings settings)
@@ -203,23 +198,20 @@ namespace Ombi.Notifications.Agents
                 Message = parsed.Message,
             };
 
-            ShowCompactEmbed.TryGetValue(NotificationType.RequestDeclined, out var compactEmbed);
-            MentionAlias.TryGetValue(NotificationType.RequestDeclined, out var mentionUser);
-
             DiscordAuthor author = new DiscordAuthor
             {
-                name = "Request Declined!",
+                name = parsed.Subject,
                 icon_url = "https://i.imgur.com/i1X39I2.png"
             };
 
             DiscordEmbeds embed = null;
             if (model.RequestType == RequestType.Movie)
             {
-                embed = createDiscordEmbed(author, MovieRequest, parsed.Image, compactEmbed, mentionUser);
+                embed = CreateDiscordEmbed(author, MovieRequest, NotificationType.RequestDeclined, parsed.Image);
             }
             else if (model.RequestType == RequestType.TvShow)
             {
-                embed = createDiscordEmbed(author, TvRequest, parsed.Image, compactEmbed, mentionUser);
+                embed = CreateDiscordEmbed(author, TvRequest, NotificationType.RequestDeclined, parsed.Image);
             }
             await Send(notification, settings, embed);
         }
@@ -237,23 +229,20 @@ namespace Ombi.Notifications.Agents
                 Message = parsed.Message,
             };
 
-            ShowCompactEmbed.TryGetValue(NotificationType.RequestApproved, out var compactEmbed);
-            MentionAlias.TryGetValue(NotificationType.RequestApproved, out var mentionUser);
-
             DiscordAuthor author = new DiscordAuthor
             {
-                name = "Request Approved!",
+                name = parsed.Subject,
                 icon_url = "https://i.imgur.com/sodXDGW.png"
             };
 
             DiscordEmbeds embed = null;
             if (model.RequestType == RequestType.Movie)
             {
-                embed = createDiscordEmbed(author, MovieRequest, parsed.Image, compactEmbed, mentionUser);
+                embed = CreateDiscordEmbed(author, MovieRequest, NotificationType.RequestApproved, parsed.Image);
             }
             else if (model.RequestType == RequestType.TvShow)
             {
-                embed = createDiscordEmbed(author, TvRequest, parsed.Image, compactEmbed, mentionUser);
+                embed = CreateDiscordEmbed(author, TvRequest, NotificationType.RequestApproved, parsed.Image);
             }
             await Send(notification, settings, embed);
         }
@@ -271,31 +260,29 @@ namespace Ombi.Notifications.Agents
                 Message = parsed.Message,
             };
 
-            // TODO implement plex / emby url
+            // TODO: implement plex / emby url
             string authorUrl = null;
             /*
             if (Customization.ApplicationUrl.HasValue())
                 authorUrl = $"{Customization.ApplicationUrl}requests";
             */
-
-            ShowCompactEmbed.TryGetValue(NotificationType.RequestAvailable, out var compact);
-            MentionAlias.TryGetValue(NotificationType.RequestAvailable, out var mentionUser);
+            
 
             DiscordAuthor author = new DiscordAuthor
             {
-                name = "Request Now Available!",
+                name = parsed.Subject,
                 url = authorUrl,
                 icon_url = "https://i.imgur.com/k4bX9KM.png"
             };
-
+            
             DiscordEmbeds embed = null;
             if (model.RequestType == RequestType.Movie)
             {
-                embed = createDiscordEmbed(author, MovieRequest, parsed.Image, compact, mentionUser);
+                embed = CreateDiscordEmbed(author, MovieRequest, NotificationType.RequestAvailable, parsed.Image);
             }
             else if (model.RequestType == RequestType.TvShow)
             {
-                embed = createDiscordEmbed(author, TvRequest, parsed.Image, compact, mentionUser);
+                embed = CreateDiscordEmbed(author, TvRequest, NotificationType.RequestAvailable, parsed.Image);
             }
 
             await Send(notification, settings, embed);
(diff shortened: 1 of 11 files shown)


src/Ombi.Notifications/Agents/DiscordNotification.cs
```diff
@@ -18,6 +18,14 @@ namespace Ombi.Notifications.Agents
 {
     public class DiscordNotification : BaseNotification<DiscordNotificationSettings>, IDiscordNotification
     {
+        // constants I needed but could not find
+        public const string IMDB_BASE_URL = "http://www.imdb.com/title/";
```
tidusjar: Can we out these after the constructor please

src/Ombi.Notifications/Agents/DiscordNotification.cs
```diff
@@ -67,7 +75,39 @@ protected override bool ValidateConfiguration(DiscordNotificationSettings settin
                 Message = parsed.Message,
             };
 
+            string userString = TvRequest.RequestedUser.Alias;
```
tidusjar: TV request could be null here if it's a movie request

src/Ombi.Notifications/Agents/DiscordNotification.cs
```diff
@@ -67,7 +75,39 @@ protected override bool ValidateConfiguration(DiscordNotificationSettings settin
...
             notification.Other.Add("image", parsed.Image);
+            if (Customization.ApplicationUrl.HasValue()) notification.Other.Add("authorUrl", Customization.ApplicationUrl + "requests");
```
tidusjar: String interpolation would be nicer than string concatenation here

src/Ombi.Notifications/Agents/DiscordNotification.cs
```diff
@@ -67,7 +75,39 @@ protected override bool ValidateConfiguration(DiscordNotificationSettings settin
...
+
+            if (model.RequestType == RequestType.Movie)
+            {
+                if (mentionAlias)
+                {
+                    notification.Other.Add("mention", userString);
+                    userString = TvRequest.RequestedUser.UserName;
```
tidusjar: This should be MoviesRequest right? Since we are checking if it's a movie above

src/Ombi.Notifications/Agents/DiscordNotification.cs
```diff
@@ -67,7 +75,39 @@ protected override bool ValidateConfiguration(DiscordNotificationSettings settin
...
+                }
+                notification.Other.Add("footer", "Requested by " + MovieRequest.RequestedUser + " on " + MovieRequest.RequestedDate.ToLongDateString());
```
tidusjar: Same regarding string interpolation here

src/Ombi.Notifications/Agents/DiscordNotification.cs
```diff
@@ -67,7 +75,39 @@ protected override bool ValidateConfiguration(DiscordNotificationSettings settin
...
+                notification.Other.Add("author", "🎬 New Movie Request!");
+                notification.Other.Add("title", MovieRequest.Title + " (" + MovieRequest.ReleaseDate.Year + ")");
+                notification.Other.Add("description", MovieRequest.Overview);
+                notification.Other.Add("titleUrl", IMDB_BASE_URL + MovieRequest.ImdbId);
+                Logger.LogDebug(MovieRequest.Background);
```
tidusjar: Why are we logging the background? Might show strange in the logs with no context

src/Ombi.Notifications/Agents/DiscordNotification.cs
```diff
@@ -200,25 +240,70 @@ protected override bool ValidateConfiguration(DiscordNotificationSettings settin
             {
                 var discordBody = new DiscordWebhookBody
                 {
-                    content = model.Message,
+                    // content = model.Message,
                     username = settings.Username,
                 };
 
+                string author;
+                model.Other.TryGetValue("author", out author);
```
tidusjar: We can do `out var author` that way we don't need to declare it above

src/Ombi.Notifications/Agents/DiscordNotification.cs
```diff
@@ -27,6 +27,14 @@ public class DiscordNotification : BaseNotification<DiscordNotificationSettings>
             Logger = log;
         }
 
+        // constants I needed but could not find
+        public const string IMDB_BASE_URL = "http://www.imdb.com/title/";
+        public const string TVDB_BASE_URL = "https://www.thetvdb.com/?tab=series&id=";
+
+        // if true mentionAlias will post the alias to discord and trigger an @mention if set up as <@id>.  
+        // It will also use the username instead of alias when talking about a user. e.g. "Requested by username on 16 March"
+        public bool mentionAlias { get; set; } = true;
```
tidusjar: Will this be settable from the UI?

tidusjar: Also since it's a public property it should have a capital M

roanvonb: Yes, the idea is to eventually have it be settable via UI once I figure out how to add that. For now now it's true by default.

src/Ombi.Notifications/Agents/DiscordNotification.cs
```diff
@@ -200,25 +242,55 @@ protected override bool ValidateConfiguration(DiscordNotificationSettings settin
             {
                 var discordBody = new DiscordWebhookBody
                 {
-                    content = model.Message,
+                    // content = model.Message,
```
roanvonb: Just realised this will break all the other notifications I haven't implemented yet. Working on a fix.

src/Ombi.Notifications/Agents/DiscordNotification.cs
```diff
@@ -126,6 +126,8 @@ protected override bool ValidateConfiguration(DiscordNotificationSettings settin
                 Message = parsed.Message,
             };
             notification.Other.Add("image", parsed.Image);
+            notification.Other.Add("author", "New Issue!");
```
tidusjar: This should be the subject, located on `notification.Subject`

src/Ombi.Notifications/Agents/DiscordNotification.cs
```diff
@@ -21,12 +22,34 @@ public class DiscordNotification : BaseNotification<DiscordNotificationSettings>
...
                                    ILogger<DiscordNotification> log, INotificationTemplatesRepository r,
                                    IMovieRequestRepository m, ITvRequestRepository t, ISettingsService<CustomizationSettings> s)
-            : base(sn, r, m, t,s,log)
+            : base(sn, r, m, t, s, log)
         {
             Api = api;
             Logger = log;
+            ShowCompactEmbed = new Dictionary<NotificationType, bool>();
+            // Temporary defaults
+            ShowCompactEmbed.Add(NotificationType.NewRequest, true);
```
tidusjar: All of these should be driven from the UI

src/Ombi.Notifications/Agents/DiscordNotification.cs
```diff
@@ -21,12 +22,34 @@ public class DiscordNotification : BaseNotification<DiscordNotificationSettings>
...
+            ShowCompactEmbed.Add(NotificationType.RequestApproved, true);
+            ShowCompactEmbed.Add(NotificationType.RequestDeclined, true);
+
+            MentionAlias = new Dictionary<NotificationType, bool>();
+            // Temporary defaults
+            MentionAlias.Add(NotificationType.RequestAvailable, true);
```
tidusjar: Same as the above here

src/Ombi.Notifications/Agents/DiscordNotification.cs
```diff
@@ -67,8 +90,30 @@ protected override bool ValidateConfiguration(DiscordNotificationSettings settin
...
+            string authorUrl = null;
+            if (Customization.ApplicationUrl.HasValue())
+                authorUrl = $"{Customization.ApplicationUrl}requests";
+
+            ShowCompactEmbed.TryGetValue(NotificationType.NewRequest, out var compactEmbed);
+            MentionAlias.TryGetValue(NotificationType.NewRequest, out var mentionUser);
+
+            DiscordAuthor author = new DiscordAuthor
+            {
+                name = "New Request!",
```
tidusjar: This should be the subject

src/Ombi.Notifications/Agents/DiscordNotification.cs
```diff
@@ -236,5 +361,106 @@ protected override bool ValidateConfiguration(DiscordNotificationSettings settin
             };
             await Send(notification, settings);
         }
+
+        private DiscordEmbed createDiscordEmbed(DiscordAuthor author, MovieRequests req, string imageUrl, bool compactEmbed, bool mentionUser)
```
tidusjar: Methods should be CamelCase

src/Ombi.Notifications/Agents/DiscordNotification.cs
```diff
@@ -236,5 +361,106 @@ protected override bool ValidateConfiguration(DiscordNotificationSettings settin
...
+            else
+                image = new DiscordImage { url = imageUrl };
+
+            string description = null;
+            if (!compactEmbed)
+                description = MovieRequest.Overview;
+
+            List<DiscordField> fields = new List<DiscordField>();
+
+            string alias = MovieRequest.RequestedUser.Alias;
```
tidusjar: The user might not have an alias set

src/Ombi.Notifications/Agents/DiscordNotification.cs
```diff
@@ -236,5 +361,106 @@ protected override bool ValidateConfiguration(DiscordNotificationSettings settin
...
+                thumbnail = thumbnail,
+                image = image,
+                description = description,
+                footer = footer,
+                fields = fields
+            };
+            return embed;
+        }
+
+        private DiscordEmbed createDiscordEmbed(DiscordAuthor author, ChildRequests req, string imageUrl, bool compactEmbed, bool mentionUser)
```
tidusjar: Since this is essentially the same as the method above but we are passing in slightly different objects, we should make a common object and populate that then pass it in. 

This way we can have 1 method. 

The new class would have the release date, overview, title properties etc.

src/Ombi.Notifications/Agents/DiscordNotification.cs
roanvonb: My bad, I changed how that worked, and then changed it back to keep the original functionality (but clearly did it wrong). On it.

src/Ombi.Notifications/Agents/DiscordNotification.cs
roanvonb: Implemented all the other changes except this as I'm not sure how to do implement that.

src/Ombi.Notifications/Agents/DiscordNotification.cs
```diff
@@ -190,8 +259,53 @@ protected override bool ValidateConfiguration(DiscordNotificationSettings settin
...
+            /*
+            if (Customization.ApplicationUrl.HasValue())
+                authorUrl = $"{Customization.ApplicationUrl}requests";
+            */
+            
+
+            DiscordAuthor author = new DiscordAuthor
+            {
+                name = parsed.Subject,
+                url = authorUrl,
```
tidusjar: authorUrl is always null here

src/Ombi.Notifications/Agents/DiscordNotification.cs
```diff
@@ -21,12 +22,34 @@ public class DiscordNotification : BaseNotification<DiscordNotificationSettings>
...
+        // constants I needed but could not find
+        public const string IMDB_BASE_URL = "http://www.imdb.com/title/";
+        public const string TVDB_BASE_URL = "https://www.thetvdb.com/?tab=series&id=";
+
+        // Whether or not to show a compact embed notification (thumbnail + no description)
+        public Dictionary<NotificationType, bool> ShowCompactEmbed;
+        // Whether or not to post the alias to discord (for each notification type) which trigger an @mention if alias is set up as <@id>.  
+        public Dictionary<NotificationType, bool> MentionAlias;
+        // Whether to use alias or username when referencing a user
+        public bool UsingAliasAsMention = true;
```
tidusjar: What's the difference between `UsingAliasAsMention` and `MentionAlias`?

tidusjar: Thinking about it i'm not sure if this option is needed since the in the body the user could just enter `@<{Alias}>`. What do you think?

tidusjar:
Ok, UI should be present now. I just dont understand why the Mention Alias thing is there?

roanvonb:
Thanks for the UI! I have some time tomorrow to check it out and fix any problems. I'll also have a look at implementing the "Issues" notifications. I'll give the UI for that a go myself based on your example for requests.

As for MentionAlias, it is whether to mention users for specific types of requests. For example, if someone requests a movie, they already know they requested it and might not want to be pinged. Or if users don't want to be pinged when their requests are approved, but do want to be pinged when their requests are available.

UsingAliasAsMention toggles off the whole system so no one will be mentioned regardless of MentionAlias setting. I agree that this is a bit superfluous and I'll remove that as you can just toggle each of MentionAlias to false.

tidusjar:
There should be no UI changes to be made going forward, I think i've done it all.

Regarding `MentionAlias`, I don't think it's needed since we have the ability to disable certain templates e.g. we can turn off the NewRequest template so that notification will not send for that agent.

Maybe I still don't understand the use for it, but I cannot see what benifit it brings?

tidusjar:
Ok I understand, but wouldn't you just edit the Request Available template to include the @<discordId> ? It seems a bit pointless adding that option when it can be done in the message if you follow me.

tidusjar:
Cool, added the Database migrations so it should all work!

tidusjar:
Hey. Is this ready from your end?

roanvonb:
Hi, sorry but I'm swamped at the moment and haven't had a look yet. New job + uni assignments keeping me down. I'll see if I can make some time in a couple of weeks, but feel free to continue without me if need to release it sooner. I'll jump back in when I can.

roanvonb:
Whoops, accidentally closed it.

goldenpipes:
is this done? what happened?

Issue #1887, opened 2018-01-12
[LC] Requests Page


leram84:
Just to get us started cause i know ur gonna want an issue, but don't worry about this until after we finish the search page stuff.

But right off the bat im gonna need a few things:

1) Can we get a "View in Plex" button for these items? No reason that should only exist in search imo.

2) TV section seems incomplete? It's missing a bunch of fields that movies have. (Request Status, Request Date, Requested By) though i guess ur putting some of that in the "view" drill down area, but it doesn't make sense to do so imo. would be better for that to be uniform w movies if possible. Ideally, the only thing that "view" button should display is the Seasons spreadsheet... everything else should be on the banner to begin w (like movies)

3) That view button (on tv items) needs an active state. Currently the css changes its background when in focus, but thats not really the right way to do a toggle button... should have an on off state instead.

4) I need id's on each button in the far right of each media item as well as an id for the wrapper (the one around ALL the buttons on the right). And please make sure to add the individual button id's at the highest level for said button... so for something like:

```
<div class="btn-group btn-split ng-star-inserted">
                            <button class="btn btn-sm btn-warning-outline" type="button"><i class="fa fa-plus"></i> Change Root Folder</button>
                            <button aria-expanded="false" aria-haspopup="true" class="btn btn-warning-outline dropdown-toggle" data-toggle="dropdown" type="button">
                                <span class="caret"></span>
                                <span class="sr-only">Toggle Dropdown</span>
                            </button>
                            <ul class="dropdown-menu">
                                <!----><li class="ng-star-inserted"><a href="#">D:\Media\Movies\</a></li><li class="ng-star-inserted"><a href="#">E:\Media\Movies\</a></li><li class="ng-star-inserted"><a href="#">F:\Media\Stand-Up Comedy</a></li>
                            </ul>
                        </div>
```
make sure to add the id to that first div, and not the individual buttons.

Ill probably need more when i really get into it, but that should let me get started on the requests page :)

goldenpipes:
[ITS HAPPENING](http://www.reactiongifs.com/r/2013/05/Ron-Paul_Its-Happening1.gif)

tidusjar:
Regarding no. 2, that is not possible. 

Since you could have multiple requests for the same series by two different people with different statues, it wouldn't work.

tidusjar:
[...]

Issue #1773, opened 2017-12-04
[LC] Need classes for media tags


leram84:
I need classes for each of those tags  (release date, home page., processing request). I can't use nth child on all those cause the number of tags changes depending on if  a movie has a homepage, etc... so i need each child (either attribute or span... doesn't matter, just need the first one) of that highlighted div to have a unique class.

![ombi tags](https://user-images.githubusercontent.com/21117493/33573081-cae5a530-d902-11e7-9667-cb4ccba8f8c4.PNG)

leram84:
OHHH and i also need those classes to be different when the text in the label changes... so one class for "processing request" and a different one for "not requested" etc

tidusjar:
I gave them some id's e.g. `processingRequestLabel`

leram84:
i dont think this is working.. i see the labels you're trying to add in that commit, but on my side they all turn into `ng-star-inserted` class... its everywhere. I need each tag to have a unique class, and the last one (the one that says Available/Not Requested/Processing Request/etc) I need a class for each possible state that tag can have... as a side note, i think the best way to do that one would be:

Not Requested
Pending Approval
Pending (for when its approved, but not added for whatever reason... including "it hasnt been released yet)
Rejected
Available

lmk if u want to see what im doing if u need convincing that this is a good idea... it makes more sense when u see the way im laying things out

tidusjar:
I added Id's:
![image](https://user-images.githubusercontent.com/6642220/34061826-aeccb2fc-e1e1-11e7-81f9-660a5d1b5aa1.png)

tidusjar:
Fixed in develop

PR #2118, opened 2018-04-01
LC requirements

anojht:
Adding wrappers and classes to UI for LC requirements. Implemented use of events when certain UI elements are clicked so can toggle active class styling. Fixed issues #1887 and #1773


src/Ombi/ClientApp/app/requests/movierequests.component.html
```diff
@@ -14,32 +14,32 @@
...
-                            <a (click)="setOrder('requestedUser.userAlias')">{{ 'Requests.SortRequestedBy' | translate }}
+                            <a (click)="setOrder('requestedUser.userAlias', $event)">{{ 'Requests.SortRequestedBy' | translate }}
                                     <span *ngIf="order === 'requestedUser.userAlias'">
                                             <span [hidden]="reverse"><small><i class="fa fa-arrow-down" aria-hidden="true"></i></small></span>
                                             <span [hidden]="!reverse"><small><i class="fa fa-arrow-up" aria-hidden="true"></i></small></span>
                                             </span>
                                         </a>
 
-                            <a (click)="setOrder('status')">{{ 'Requests.SortStatus' | translate }}
+                            <a (click)="setOrder('status', $event)">{{ 'Requests.SortStatus' | translate }}
```
anojht: Passing the click event to the setOrder function so that we can use it to do active style CSS  class toggling based on the currently selected sort filter.

src/Ombi/ClientApp/app/requests/movierequests.component.html
```diff
@@ -78,40 +78,42 @@ <h4 class="request-title">{{request.title}} ({{request.releaseDate | date: 'yyyy
...
-                    <span>{{ 'Requests.RequestedBy' | translate }} </span>
-                    <span *ngIf="!isAdmin">{{request.requestedUser.userName}}</span>
-                    <span *ngIf="isAdmin && request.requestedUser.alias">{{request.requestedUser.alias}}</span>
-                    <span *ngIf="isAdmin && !request.requestedUser.alias">{{request.requestedUser.userName}}</span>
-                </div>
-                <div>
-                    <span>{{ 'Requests.Status' | translate }} </span>
-                    <span class="label label-success" id="requestedStatusLabel">{{request.status}}</span>
-                </div>
+                <div class="request-info">
```
anojht: Created a wrapper to enclose the request info for movies and added a class to each item label

src/Ombi/ClientApp/app/requests/movierequests.component.ts
```diff
@@ -175,11 +175,18 @@ export class MovieRequestsComponent implements OnInit {
...
+
+        const parent = el.parentElement;
+        const previousFilter = parent.querySelector(".active");
+
         if (this.order === value) {
-          this.reverse = !this.reverse;
+            this.reverse = !this.reverse;
+        } else {
+            previousFilter.className = "";
+            el.className = "active";
```
anojht: This function receives the event as a parameter gets the element that the click occurred from and the parent element. We query for whether an element is active and if its not the element where the click occurred we remove the active class from the default or previous element and enable the active class on the element where click occurred.  This is for styling currently active filter in the movie requests filter option.

src/Ombi/ClientApp/app/requests/tvrequests.component.ts
```diff
@@ -66,7 +66,19 @@ export class TvRequestsComponent implements OnInit {
...
+
+            const isButtonAlreadyActive = el.parentElement.querySelector(".active");
+            // if a Button already has Class: .active
+            if (isButtonAlreadyActive) {
+                isButtonAlreadyActive.classList.remove("active");
+            } else {
+                el.className += " active";
+            }
+        }
+
```
anojht: Similar idea to what is done above for movie requests component but instead we apply the active styling to the View button of the TV shows request page since current behavior is not correct in that when the button is toggled and clicked away it stays highlighted as if active.

src/Ombi/ClientApp/app/usermanagement/usermanagement-edit.component.html
```diff
@@ -1,68 +1,70 @@
 ﻿<div *ngIf="user">
-    <h3>User: {{user.userName}}</h3>
-    <button type="button" class="btn btn-primary-outline" style="float:right;" [routerLink]="['/usermanagement/']">Back</button>
+    <div class="user-details">
```
anojht: Created another wrapper to enclose the entire edit user page since by default the ID that was being assigned to the page was a generated one from the routed component.

anojht (commented):
Added comments everywhere I could!

tidusjar:
Thanks I'll take a look when I can!

tidusjar:
Thanks!

Issue #2124, opened 2018-04-05
[LC] Loose ends


leram84:
I'm sure there will be more "requests", but im fairly confident this is the last set of issues i could use before i actually release omBlur! @anojht has already done most of these (THANK YOU!!), but wanted to put it all down so we're all on the same page, and you can yell at us if any of this is out of line lol:

1)  The "report issue" button on the tv requests page is buried in the "view" drop down button. I need that to be on the same level as the view button/root folder/quality profile buttons, which is how it is for non admin users. And it makes sense since that is a "show wide" feature anyway. 

2) Would it be possible to add a NEW "remove" button on that level as well? movies request page has the "remove" button there, but in tv its tied to each specific request... so if 2 people request different eps of a show, you can remove each persons request one by one... thats great, but also having a remove button on the tile itself before you click the "view" dropdown would be a fast way to just clear ALL requests for that show (and would make it look like movies). 

3) The new issues details page has an id... but the id has the issue number in it, so basically theres no way to use it as an id... I need the id to be generated in a way that makes it the same for all issues.

4) 2 things from the last pr... i see the "active" classes @anojht added for the headers on the user management page and the sort dropdown on the requests page, but not for the headers on the "issues" page  or the "filter" dropdown on the requests page.

5) Can we update font-awesome and also add glyphicon-halflings to ombi? There are a few glyphs i want to use and a bunch of font-awesome ones that aren't included in whatever the current version ombi has.

6) Can we add a default poster to every item throughout ombi so if no poster loads it will show the default at least instead? `@odrel` has generously created these:
Movies: https://cdn.discordapp.com/attachments/374757488005873667/431539170268282881/Unknown-movie_notransparency.png
TV: https://cdn.discordapp.com/attachments/374757488005873667/431539008791904267/Unknown-tv_notransparency.png

7) Ombi returns a lot of junk result (just a description, and no other info, or just the release date and nothing else)... i think it would be good to have a condition that if there is no title, then dont show the result. Theres no point in requesting something with no title so would look a lot cleaner to just remove those results. If that's too much, Adding a placeholder title like `No Title` would be very helpful in keeping the css from getting all messed up when the title is absent (same issue i was having w missing posters)

[...]

PR #2125, opened 2018-04-05
Knocking out LC requirements in issue #2124

anojht:
This PR is a WIP and will serve the changes required for LC to reach completion.


src/Ombi/ClientApp/app/requests/tvrequests.component.ts
```diff
@@ -223,6 +222,22 @@ export class TvRequestsComponent implements OnInit {
         this.loadInit();
     }
 
+    private setDefaults(val: any) {
+        if (val.data.posterPath === null) {
+            val.data.posterPath = "../../../images/default_tv_poster.png";
+        }
+
+        if (val.data.imdbId.length === 0) {
+            this.searchService.getShowInformationTreeNode(val.data.tvDbId).subscribe(x => {
```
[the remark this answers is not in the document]
anojht: In search results the TreeNode data has `seriesId` attribute which can be used with the url: `https://www.tvmaze.com/shows/{seriesId}` in order to display titles with urls in the event that that tvmaze data does not contain an `imdbId` for the particular show. 

However inside the requests page the TreeNode data does not have a `seriesId` attribute, so the only way I thought of to get this since we do have the `tvDbId` is to use the search service call to obtain this very reliably and this way tv show titles don't just link to a 404 imdb page. 

What do you think is a better approach? I mean we can set the `imdbId` attribute accordingly during search so we don't have to do this and this would work for new requests but then I was just thinking of the existing requests...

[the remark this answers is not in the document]
tidusjar: Well we do try and set the IMDb Id, but we can't always get it because it's not available sometimes. What I'd recommend is if we dont have it then we either create a new api and get it OR get it on the requests API that returns the requested items.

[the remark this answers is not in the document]
tidusjar: That's fine! We just need a new API assembly like all the other external API's.

Should be pretty simple to do.

src/Ombi.Core/Helpers/TvShowRequestBuilder.cs
```diff
@@ -16,23 +18,35 @@ namespace Ombi.Core.Helpers
...
 
         public async Task<TvShowRequestBuilder> GetShowInfo(int id)
         {
             ShowInfo = await TvApi.ShowLookupByTheTvDbId(id);
+            Result = await MovieDbApi.SearchTv(ShowInfo.name);
+            foreach (TvSearchResult result in Result) {
+                if (result.Name == ShowInfo.name)
+                {                  
+                    var showIds = await MovieDbApi.GetTvExternals(result.Id);
+                    ShowInfo.externals.imdb = showIds.imdb_id;
```
tidusjar: Since we are here we can store `TheMovieDbId` too.

src/Ombi/ClientApp/app/requests/tvrequests.component.ts
```diff
@@ -227,15 +226,7 @@ export class TvRequestsComponent implements OnInit {
...
-        if (val.data.imdbId.length === 0) {
-            this.searchService.getShowInformationTreeNode(val.data.tvDbId).subscribe(x => {
-                    if (x.data) {
-                        val.data.imdbId = "https://www.tvmaze.com/shows/" + x.data.seriesId;
-                    }
-            });   
-        } else {
-            val.data.imdbId = "http://www.imdb.com/title/" + val.data.imdbId + "/";
-        }
+        val.data.imdbId = "http://www.imdb.com/title/" + val.data.imdbId + "/";
```
tidusjar: This is not really the `imdbId`. I assume we use this on the UI. I'd prefer if we just do this in the UI

tidusjar (changes requested):
Nice job!

src/Ombi.Core/Helpers/TvShowRequestBuilder.cs
anojht: `ShowInfo` here is of type `TvMazeShow` and it does not have `TheMovieDbId` as one of its attributes. Shall I extend the type to include this and then I can set it here?

anojht: Wrote a DB migration and extended the TvRequests to include `TheMovieDbId` and it seems to be working beautifully. Is this ok to do? If so I will push the commit.

src/Ombi/ClientApp/app/requests/movierequests.component.ts
```diff
@@ -149,15 +149,40 @@ export class MovieRequestsComponent implements OnInit {
...
         this.filter.statusFilter = FilterType.None;
         
         this.resetSearch();
     }
 
-    public filterAvailability(filter: FilterType) {
+    public filterAvailability(filter: FilterType, el: any) {
+        el = el.toElement || el.relatedTarget || el.target || el.srcElement;
+
+        el = el.parentElement; //gets radio div
```
tidusjar: Can we not move this part of the code into it's own private method? Since it's used below too

src/Ombi/ClientApp/app/requests/movierequests.component.ts
```diff
@@ -166,7 +191,23 @@ export class MovieRequestsComponent implements OnInit {
...
 
-    public filterStatus(filter: FilterType) {
+    public filterStatus(filter: FilterType, el: any) {
+        el = el.toElement || el.relatedTarget || el.target || el.srcElement;
+
+        el = el.parentElement; //gets radio div
+        el = el.parentElement; //gets form group div
+        el = el.parentElement; //gets status filter div
+        el = el.querySelectorAll("INPUT");
+        for (el of el) {
```
tidusjar: See above

src/Ombi/ClientApp/app/search/moviesearch.component.ts
```diff
@@ -157,12 +157,15 @@ export class MovieSearchComponent implements OnInit {
...
-           
-            val.background = this.sanitizer.
-            bypassSecurityTrustStyle
-            ("url(" + "https://image.tmdb.org/t/p/w1280" + val.backdropPath + ")");
-            this.searchService.getMovieInformation(val.id)
+       this.movieResults.forEach((val, index) => {
+           if (val.posterPath === null) {
+               val.posterPath = "../../../images/default_movie_poster.png";
+           } else {
+               val.posterPath = "https://image.tmdb.org/t/p/w300/" + val.posterPath;
```
tidusjar: Again can we put the concatenation in the UI?

src/Ombi/ClientApp/app/requests/movierequests.component.ts
```diff
@@ -296,11 +338,20 @@ export class MovieRequestsComponent implements OnInit {
...
         this.setBackground(req);
         this.setQualityOverrides(req);
         this.setRootFolderOverrides(req);
     }
 
+    private setPoster(req: IMovieRequests): void {
+        if (req.posterPath === null) {
+            req.posterPath = "../../../images/default_movie_poster.png";
+        } else {
+            req.posterPath = "https://image.tmdb.org/t/p/w300/" + req.posterPath;
```
tidusjar: Can this go in the HTML, please?

tidusjar (changes requested):

anojht (commented):
This PR is done 😄  Please review!

tidusjar (approved):

PR #2161, opened 2018-04-15
Add web-app-capable for IOS and Android

thomasvt1:
Add support for both IOS and Android.

Feathub: https://feathub.com/tidusjar/Ombi/+270


tidusjar:
Awesome thanks!

PR #2172, opened 2018-04-19
added background property to tvrequests API

anojht:
See discord for more details, @causefx requested this as tvrequests does not have this property but movierequests do.


tidusjar:
Thanks!

Issue #2177, opened 2018-04-20
Themes broke


si0972:
#### Ombi build Version:

3.0.3203

#### Update Branch:

Develop

#### Operating System:

Docker

#### Problem Description:

#2148 actualy broke it. I get 500 errors in console, and a error when i browse the link manualy .

Url: `domain.com/ombi/api/v1/Settings/ThemeContent?url=https://raw.githubusercontent.com/leram84/layer.Cake/master/ombi/css/omBlur.css`
Error:`{"error":"Specified argument was out of the range of valid values.\nParameter name: startIndex"}`
Base: `/ombi`

#### Reproduction Steps:

update ombi to dev branch

tidusjar:
This should be fixed in develop

PR #2180, opened 2018-04-20
fixing baseurl breaking themes issue #2177

anojht:
see issue #2177


Unimatrix0:
This breaks installs using custom baseurls.

Unimatrix0:
Getting `{"error":"Specified argument was out of the range of valid values.\nParameter name: startIndex"}`

anojht:
So it turns out if you update with the theme already set. This code will keep injecting base urls in front thus causing updated installations to become unresponsive. Thanks to @Unimatrix0 and @goldenpipes for testing this. I am submitting a new PR that does a check if baseurl is already present in the theme url and then remove and reset the theme url. Just to cover cases that have occures such as `/ombi/ombi/ombi.....`

PR #2225, opened 2018-05-04
Fixed duplicate release years in titles for Plex newsletter

anojht:


src/Ombi.Schedule/Jobs/Ombi/NewsletterJob.cs
```diff
@@ -547,7 +547,7 @@ private void CreateMovieHtmlContent(StringBuilder sb, MovieResponseDto info, str
                     AddMediaServerUrl(sb, t.Url, banner);
                     AddInfoTable(sb);
 
-                    var title = $"{t.Title} ({t.ReleaseYear})";
+                    var title = $"{t.Title} ({info.premiered.Remove(4)})";
```
tidusjar: We need to make sure there are > 4 items in there.

So why does this fix it?

We could also take it a step further and not include the `()` if there is no date?

anojht: `info.premiered` returns date formatted as `2018-05-04` so `.Remove(4)` trims after the year. This is the same method that is done for Emby and since I use Emby that's what I tested with.

However you are right in that a check needs to be performed in order to account for when TVMaze does not have the date for the TV show.

PR #2260, opened 2018-05-19
Sort TvRequests by latest request

errorhandler:
This small change makes the TV request list sorting a bit more intuitive by ordering by the request date in descending order. This makes approving newly requested shows much easier.


tidusjar:
Thanks!

PR #2294, opened 2018-06-01
Use selected episodes in submitRequest

CalvinWalzel:
Currently it's not possible to edit an already requested series, because the request params are populated with already requested episodes in addition to the selected ones. I hope this will fix it.


tidusjar:
Thanks!

PR #2332, opened 2018-06-17
[WIP] Update to Angular 6/Webpack 4

MattJeanes:
Do not merge yet, still has some issues/WIP

- Rebased onto 6.0.0 AngularBasic template
  - Upgraded to Angular 6
  - Upgraded to Webpack 4
  - Completely reworked webpack bundling
  - Use AOT in production
  - Better bundle splitting/optimisation
  - No longer show sourcemaps in prod builds (but you can change this however you like, devtool option in webpack.config.ts)
  - Much smaller bundles as a result of AOT/Angular/Webpack upgrade improvements
- Rxjs syntax has been upgraded to v6
- Removed import file as no longer needed under Rxjs v6
- All packages upgraded on npm, with exception of bootstrap/bootswatch as 4.0 has big breaking changes
- Changed some code to fix AOT build


MattJeanes:
Currently an issue with prod builds, something to do with rxjs operators/chunk splitting.. Running locally works fine if you want to test it though. @tidusjar can I get some test apis so I can actually use the app properly? Hard to test without it. Everything kept 500'ing as (I assume) I hadn't set up any of the sonarr/plex api urls etc.

MattJeanes:
@tidusjar prod build issue fixed - TreeTable still nobbed but no errors, can't even get it to display anything using their own official example on a very basic angular app. Will investigate, but getting much closer to done.

tidusjar:
Thanks, I'll going to be removing the dependency on the tree table. Don't really need it, can rewrite it with basic html.

PR #2357, opened 2018-06-27
Minor style tweaks

randybruder:
I have a custom style sheet on my Ombi install that has a bunch of changes, but I realized some of my tweaks would be useful to more people. Here's some minor tweaks I've made and explanations why:

---

```css
.backdrop{
    background-position: 50% 33%;
}
```
Almost always, the backgrounds from TMDb have the focus of the image in the top half. This often results in the background images showing heads cut off. Positioning the background image higher within the container tends to make a much better cropping.

---

```css
.checkbox label:before {
    top: 3px;
}
```
If the label next to the checkbox breaks over two lines, then the "fake" checkbox ends up inline with the second line of text, rather than the first. Specifying a top value as well corrects this.

---

```css
.tags {
    display: block;
}
```
The paragraph following the tags is very close, and could use a little space between the two elements. Making the tags container display as a block fixes the spacing issue.

---

```css
.btn {
    margin-bottom: 10px;
}
```
Likewise, the buttons are all immediately next to each other. Adding a bottom margin spaces them out a little, and hopefully looks better.


goldenpipes:
Excellent. Can u post a screenshot showing these changes as well?

Cheers!

randybruder:
Sure!

Obviously from the screenshots I have some of my own additional tweaks (like using the Plex brand webfont instead.)

This shows three of the tweaks (background positioning, space between the buttons, and space between the tags and description):
image

Here's an example of the checkbox:
image

If you don't set the `top` property, here's how it looks:
image

goldenpipes:
Man that cleans up real nice!

tidusjar:
Nice job!

Thanks for these changes :)

Issue #2464, opened 2018-08-24
User can request more tv episodes than allowed


markisoke:
#### Ombi build Version:

V 3.0.3587

#### Update Branch:

Master

#### Media Sever:

Plex

#### Media Server Version:

1.13.6.5339

#### Operating System:

Ubuntu 16.04

#### Ombi Applicable Logs (from `/logs/` directory or the Admin page):

``2018-08-24 01:27:07.191 +02:00 [Error] StatusCode: BadRequest, Reason: Bad Request, RequestUri: http://api.tvmaze.com/lookup/shows?thetvdb=0
2018-08-24 01:27:07.219 +02:00 [Error] StatusCode: NotFound, Reason: Not Found, RequestUri: http://api.tvmaze.com/shows/0/episodes
2018-08-24 01:27:07.270 +02:00 [Error] Something bad happened, ErrorMiddleware caught this
Newtonsoft.Json.JsonSerializationException: Cannot deserialize the current JSON object (e.g. {"name":"value"}) into type 'System.Collections.Generic.List`1[Ombi.Api.TvMaze.Models.TvMazeEpisodes]' because the type requires a JSON array (e.g. [1,2,3]) to deserialize correctly.
To fix this error either change the JSON to a JSON array (e.g. [1,2,3]) or change the deserialized type so that it is a normal .NET type (e.g. not a primitive type like integer, not a collection type like an array or List<T>) that can be deserialized from a JSON object. JsonObjectAttribute can also be added to the type to force it to deserialize from a JSON object.
Path 'name', line 1, position 8.
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObject(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)
   at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.Deserialize(JsonReader reader, Type objectType, Boolean checkAdditionalContent)
   at Newtonsoft.Json.JsonSerializer.DeserializeInternal(JsonReader reader, Type objectType)
   at Newtonsoft.Json.JsonConvert.DeserializeObject(String value, Type type, JsonSerializerSettings settings)
   at Newtonsoft.Json.JsonConvert.DeserializeObject[T](String value, JsonSerializerSettings settings)
   at Ombi.Api.Api.Request[T](Request request) in C:\projects\requestplex\src\Ombi.Api\Api.cs:line 77
[...]

PR #2467, opened 2018-08-26
Allow users with request limits to view the amount of requests they have remaining

MrTopCat:
Search pages now display remaining requests, alongside total number of requests for users that have a request limit.

Also fixes the following bugs involving TV request limits:
* When requesting multiple TV episodes in one go, only a single TV episode would be removed from the users quota.
* When requesting an entire TV show, only the episodes in the last series would be used to calculate the remaining quota.

**Movies:**
image

**TV Shows:**
image


src/Ombi.Core/Engine/TvRequestEngine.cs
```diff
@@ -608,9 +609,49 @@ private async Task<RequestEngineResult> AfterRequest(ChildRequests model)
...
+                    Limit = 0,
+                    Remaining = 0,
+                    NextRequest = DateTime.Now,
+                };
+            }
+
+            IQueryable<RequestLog> log = _requestLog.GetAll().Where(x => x.UserId == user.Id && x.RequestType == RequestType.TvShow);
+
+            // Needed, due to a bug which would cause all episode counts to be 0
+            int zeroEpisodeCount = await log.Where(x => x.EpisodeCount == 0).Select(x => x.EpisodeCount).CountAsync();
```
tidusjar: Wait, why do we do this?

src/Ombi/ClientApp/app/requests/remainingrequests.component.html
```diff
@@ -0,0 +1,18 @@
+﻿<div *ngIf="remaining?.hasLimit">
+    <h4 id="remainingRequests" class="text-center">
+        {{remaining.remaining}}/{{remaining.limit}} requests remaining
```
tidusjar: Can we add these into the translations please? Files are in `wwwroot/translations/en.json` Same with the below

src/Ombi/ClientApp/app/services/request.service.ts
```diff
@@ -2,21 +2,43 @@ import { PlatformLocation } from "@angular/common";
...
+        return this.http.get<IRemainingRequests>(`${this.url}movie/remaining`, {headers: this.headers});
+    }
+
+    public getRemainingTvRequests(): Observable<IRemainingRequests> {
+        return this.http.get<IRemainingRequests>(`${this.url}tv/remaining`, {headers: this.headers});
+    }
+
     public requestMovie(movie: IMovieRequestModel): Observable<IRequestEngineResult> {
-        return this.http.post<IRequestEngineResult>(`${this.url}Movie/`, JSON.stringify(movie),  {headers: this.headers});
+        const observer = Observable.create(observer => {
```
tidusjar: why are we doing this?

src/Ombi/ClientApp/app/requests/remainingrequests.module.ts
```diff
@@ -0,0 +1,32 @@
+﻿import { CommonModule } from "@angular/common";
```
tidusjar: What's this module for? It doesn't seem to be used anywhere and no components are being imported into it that we have created

tidusjar:
Hey man, good idea!

Nice work! Just a few questions left in the review :) Thanks!

src/Ombi.Core/Engine/TvRequestEngine.cs
MrTopCat: See #2464 for the bug I'm referring to.

Prior to this PR, we weren't saving the `EpisodeCount` to the logs, which means that all of the old log entries will all have `EpisodeCount` set to 0. 

This query gets around this issue by counting all of these instances and assuming that they have 1 episode assigned to them.

On a side note, looks like I'm searching through all requests, rather than just requests in the past week. Will fix this and check that I haven't done it elsewhere.

src/Ombi/ClientApp/app/services/request.service.ts
MrTopCat: I needed a way of passing a message to the `remainingrequests` component after a request is made, so it could update the number of requests that it's displaying. Do you think there's a cleaner way of doing this?

tidusjar: Yeah personally i'd use an `EventEmitter` to emit a new event that the `remainingrequests` component can subscribe to. That's the usual way that data would get passed around in the frontend.

MrTopCat: Cool - I'll look into changing that.

MrTopCat:
Made some changes again. Still don't know if it's ideal, but I've at least moved that logic out of the `search` service and made it so that the components handle the logic themselves.

See https://github.com/tidusjar/Ombi/pull/2467/commits/4cad24f8e95b152358fe67619b2efa62678a8328 for the changes.

tidusjar:
Cool, One last thing, you probably can remove the other changes you made in the `RequestService` now e.g. Ln60 `Observable<IRequestsViewModel<IMovieRequests>> {` 

And also there's still that module that does nothing?

markisoke:
Maybe also add the option to view the requests left in the admin userlist as shown below (please don't mind my mad paint skills)
screenshot-ombi xena link-2018 08 28-11-10-28

tidusjar:
@markisoke Please raise that on feathub.

MrTopCat:
@tidusjar This can be looked at again.

tidusjar:
Awesome, all looks good. Everything finished from yourside?

MrTopCat:
Yup, all finished on my end.

tidusjar:
Thanks man!

PR #2496, opened 2018-09-05
Display quota on user managment page

MrTopCat:
Adds information about each users remaining tv/movie quota to the user management page.

Also fixes a bug which would prevent TV requests from being logged if the user requesting has auto approve enabled. This would effectively allow those users to request an infinite number of TV shows.

image


tidusjar:
Looks good! Thanks

PR #2533, opened 2018-09-23
Fixed: Removes Lidarr legacy command result variables

Qstick:
This removes the Sonarr legacy command result variables and changes them to their newer values. These  legacy mappings have been removed in Lidarr (and Sonarr V3 api). 

I know you guys are not using for anything yet, but just wanted to make sure there were not issues down the line.


tidusjar:
Thanks man!

Issue #2564, opened 2018-10-05
Don't show sonarr options default folder and quality profile if there is only 1


l-t-k:
#### Ombi build Version:

3.0.3795

#### Update Branch:

Master

#### Media Sever:

Plex

#### Problem Description:

Ombi shouldn't provide the user the option to change default folder and quality profile if sonarr is only using one default folder and or one quality profile.

![11](https://user-images.githubusercontent.com/3150962/46534449-6affa580-c8a8-11e8-89cd-fcaa87918d4f.jpg)

![12](https://user-images.githubusercontent.com/3150962/46534513-aa2df680-c8a8-11e8-9231-99d76d0e3857.jpg)

tidusjar:
Please raise this on feathub

l-t-k:
Why? I do not think this is a feature. A dropdown menu with only one option isn't very useful in opinion.

tidusjar:
Sorry I misread it.

tidusjar:
Fixed in develop

PR #2566, opened 2018-10-06
Hides Sonarr options on tv requests page if only 1 option available.

bernarden:
A fix for #2564.


tidusjar:
Nice job! Thanks a lot!

PR #2569, opened 2018-10-06
Switch to Matterhook Client Nuget Package

PromoFaux:
See conversation here: https://github.com/tidusjar/Ombi/commit/55265211b487adcc9b652485f24f660a5090e4c1

As far as I can tell, this is all that needs doing. It builds, but someone with a more intimate knowledge of the Ombi codebase may want to confirm.

For instance, in your own implementation of the MatterhookClient class, the `PostAsync` method took an `IApi` object as a parameter... something which the main library does not do (or need?)


PromoFaux:
Actually, thinking about this some more.. the `Ombi.API.Mattermost` dll can probably go the distance, right? Or would you rather keep it do consistency?

PromoFaux:
Sorry, by "go the distance" (poor choice of English I guess!) I mean be removed from the project. 

But looking at it more closely, it makes sense to keep it in there. I was thinking that as the propsed patch now references `Matterhook.NET.MatterhookClient` directly, there is no need for the additional class for interfacing with it.

tidusjar:
The whole separation may look slightly over the top, but it allows me to encapsulate the implementation and swap it out without having to change the other code

PromoFaux:
Yeah, that makes sense. I think what threw me was the `_api` object you were passing into your implementation of `SendAsync()`. In any case, you know the code better than I, hopefully this works!

PR #2570, opened 2018-10-06
Bug/2564 hide Radarr options when only 1 option available

bernarden:
After fixing #2564 I have noticed that Radarr has the same options available. Applying the same fix for consistency.


tidusjar:
Thanks again!

Issue #2407, opened 2018-07-28
Language selector labels should be native


Saviq:
#### Ombi build Version:

V  3.0.3511 

#### Update Branch:

Develop

#### Media Sever:

N/A

#### Operating System:

Ubuntu 18.04 container

#### Problem Description:

The language labels ("English", "French", "German" etc.) in the language selector should not be translated, instead they should be written in the native language ("English", "Français", "Deutsch", etc.), otherwise imagine clicking a language completely unknown to you, or even in an alphabet you can't read - you'd have to then do trial'n'error to find your way back.

tidusjar:
This is mostly how language selections work. E.g. change the language on
your mobile and they usually translate the languages

PR #2571, opened 2018-10-07
Changes language selector to always show native language name.

bernarden:
This PR is for issue #2407. It restricts language selector to always show native language name. 
I took all native language names from the respective translation files except for Swedish one. I believe that it should Svenska instead. Let me know if you would like this change done differently.


tidusjar:
That's excellent! Thank you very much!

Issue #2298, opened 2018-06-04
Some checkboxes not clickable in Safari


hagabla:
#### Ombi build Version:

V 3.0.3346

#### Update Branch:

stable

#### Operating System:

Server: Debian 8
Client: MacOS/Safari (latest)

#### Problem Description:

Checkboxes in the mass email settings are not clickable with Safari, only "select all" is.

#### Reproduction Steps:

Open the glorious Safari Browser, load obi website -> settings -> notifications -> mass email
Try to click the checkbox of a user.

Edit: Same issue on iOS/Safari

tidusjar:
Can you please try this in the develop branch?

hagabla:
Same issue.

tidusjar:
Can you please try this in develop now?

tidusjar:
this was fixed

PR #2581, opened 2018-10-10
Fixes untickable mass email checkboxes in Safari.

bernarden:
A fix for issues #2298. 

I have noticed that on Safari 'click' trigger was calling to selectSingleUser function after ngModel does the binding, therefore unticking the just ticked checkbox. Since that function was doing nothing other than flipping a boolean value I removed it and left this functionality up to ngModel to handle. 

P.S. Another fix that could have worked in this case is changing 'click' trigger to 'ngModelChange' one. This one seems to be consistent across the browsers.


tidusjar:
Nice job! No idea what I was thinking when I did that :P 
Thanks

PR #2612, opened 2018-10-21
Refactors search bar out of movie/tv/music requests pages into its own component.

bernarden:
Hi. I have noticed that search bar has been duplicated (html and ts) across multiple requests pages. So as a good learning exercise, I have extracted it out into its own component. 

I'm still just learning Angular, so please let me know what should be done differently.

These changes should not have affected any current behavior, except for one of issues I have noticed while testing. Filter check-boxes are not being selected properly on movies and music pages due to names and ids being the same for both pages.

A few other issues I have noticed (but not fixed) while being in this area:
1. If search returns more than 10 results and user clicks on page 2, Ombi shows simply page two as if no search has happened.
1. `Status` sorting is not supported for music requests on server side (error 500), but is an option on client side.
1. Searching ignores selected filters and sorting option.
1. `Filter` and `Sort` buttons look weird on desktop (and mobile kinda too) at btn-sm size.

### DIFF
--- a/src/Ombi/ClientApp/app/requests/movierequests.component.html
+++ b/src/Ombi/ClientApp/app/requests/movierequests.component.html
@@ -1,48 +1,9 @@
-<div class="form-group">
-    <div class="input-group">
-        <input type="text" id="search" class="form-control form-control-custom searchwidth" placeholder="Search" (keyup)="search($event)">
-        <span class="input-group-btn">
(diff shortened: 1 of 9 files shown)


tidusjar:
Hey, sorry i've not got around to looking into this yet!

I shall shortly!

tidusjar:
Sorry i'm currently not going to merge this as the search bars do different things and are quite different between each media type now. 

Thanks for the contributions though

Issue #2658, opened 2018-11-27
{Alias} notification variable does not resolve in Welcome Email sent to user.


bellington74:
#### Ombi build Version:

V 3.0.3988

#### Update Branch:

Master

#### Media Sever:

Plex

#### Media Server Version:

1.14.0.5468

#### Operating System:

Windows Server 2008 R2 Standard

#### Problem Description:

When assigning an "Alias" under User Management and then sending a Welcome Email to the configured user, the email does not pass the {Alias} variable, instead it defaults to the {UserName} variable in the **Hello {Alias}!** section.

### Reproduction Steps:

Verify that **Hello {Alias}!** is in the Welcome Email body.
Create or import a new User from Plex. (test user you have email for)
Edit the User Details to include an Alias, save.
Select "Send Welcome Email".
Check email and note that the body says "Hello username" instead of Alias.

Screenshot of User Management attached.
Snippet of Welcome Email Code attached.
Snippet of received email attached.

<img width="1146" alt="screen shot 2018-11-26 at 6 42 04 pm" src="https://user-images.githubusercontent.com/15723524/49055144-6a4e0600-f1ab-11e8-8b8e-4e2353e96d11.png">

**Code**
Hello {Alias}!<br />
You have been invited to use ********** Site!<br />
Your Username is: {UserName}<br />

**Email**
Hello testuser!
You have been invited to use ********** Site!
Your Username is: testuser

tidusjar:
Fixed in develop

PR #2659, opened 2018-11-28
Maps alias email variable for welcome emails.

bernarden:
Fix for #2658. Looks like user's alias property hasn't been mapped at controller level.


tidusjar:
Excellent! Thanks!

PR #2951, opened 2019-04-26
fix the UI a little

goldenpipes:
by moving the id="removeBtn" from the form tag to the button itself.


tidusjar:
What's the purpose of this?

goldenpipes:
current: 
https://i.imgur.com/9Ml8UkT.png

with fix: 
https://i.imgur.com/UoUAjVW.png

this allows us to use float right on the buttons to align them to the right side of the container which is a lot neater then just aligning them with a `left: 75%;`

goldenpipes:
nothing calls that form with that specific id: right? that id is used elsewhere on buttons on the TV requests page. but not on the form. just on the button.

tidusjar:
Ok cool!

PR #2956, opened 2019-04-27
re-re fix the remove button

goldenpipes:


tidusjar:
Let's not talk About this 😂

PR #2973, opened 2019-05-09
fix album-request-permission error message

Patricol:
I'm (dangerously) assuming that `Task.FromResult(Fail(message))` doesn't care about the contents of `message`.


Patricol:
Sure; I had been matching style with the rest of the file.

Patricol:
Also note that this changes the behavior when `obj.RequestType` is none of `RequestType.Movie`, `RequestType.TvShow`, or `RequestType.Album`.

tidusjar:
There are build issues with this, 
Basically if the request type is not movie, TV show or album nothing is returned. 

It's not currently possible but the compiler will not allow it. So we need to add a default return statement. Maybe just a successful return.

Patricol:
Oh shoot, congratulations on having properly setup CI!

Many projects just have perpetually failing tests; when I checked through the output and saw red-highlighted dependency errors, I assumed it was unrelated to this PR.

IMO a successful return might make odd behavior harder to diagnose. Wouldn't it be better (and more consistent with current behavior) to reject by default, with an error message?

Because developers are most likely to see that message (when adding new `RequestType`s), maybe "`Permission check failed; unknown RequestType`"? Or something more end-user friendly?

If you disagree, I can just add `return Task.FromResult(Success());` as per your suggestion.

tidusjar:
Haha no problem. 

Actually it might be better to throw some sort of custom exception. Since the application should never be in that state. Maybe something like an InvalidDataException

Patricol:
Added, though unfamiliar with C#. Syntax seems to be correct.

tidusjar:
Build is still failing, you need to import System.IO

Patricol:
That seems to have done the trick!

tidusjar:
Excellent! Thanks a lot for this :)

PR #2988, opened 2019-05-23
Swagger index prepends configured baseurl

### LINKED ISSUE #2793: Swagger index doesn't prepend configured baseurl
**Describe the bug**

The Swagger HTML landing page does not prepend the configured baseurl when trying to reach the Swagger endpoint. For instance, I am configured for a reverse proxy through nginx such that: `https://hostname/ombi/search` is the Ombi homepage when logged in. The Ombi homepage and other functionality functions normally and has expected behavior. However, when visiting `https://hostname/ombi/swagger/index.html` the webpage presented loads and the XMLHTTPRequest sent via JavaScript fails because it is trying to reach the endpoint: `https://hostname/swagger/v1/swagger.json` instead of `https://hostname/ombi/swagger/v1/swagger.json`

**To Reproduce**
- Configure Ombi to use a baseurl (/Settings/Ombi)
- Access Swagger under the baseurl (/baseurl/Swagger/index.html)
- Swagger fails to display properly

**Expected behavior**

**Screenshots**

image

image

 - Version 3.0.4119
- Media Server Plex

**Additional context**

https://github.com/tidusjar/Ombi/blob/5ee25be669e0ae7f3b9754656530197e0237731b/src/Ombi/Startup.cs#L244

au5ton:
Intended to resolve #2793 

I haven't written C# before, but I did verify the return values by looking through the Classes used and Microsoft's C# documentation.

I'm not sure how to run tests or what edge cases to look for, but I found no issues when running locally using the [build instructions](https://github.com/tidusjar/Ombi/wiki/Building-V4). I also verified that this works with the [recommended nginx setup](https://github.com/tidusjar/Ombi/wiki/Reverse-Proxy-Examples#nginx) when Swagger previously did not work through a baseurl.


src/Ombi/Startup.cs
```diff
@@ -217,7 +217,14 @@ public IServiceProvider ConfigureServices(IServiceCollection services)
             app.UseSwagger();
             app.UseSwaggerUI(c =>
             {
-                c.SwaggerEndpoint("/swagger/v1/swagger.json", "My API V1");
+                if (settings.BaseUrl.HasValue())
+                {
+                    c.SwaggerEndpoint(settings.BaseUrl + "/swagger/v1/swagger.json", "My API V1");
```
tidusjar: Excellent job! Just one thing!

Can we use string interpolation instead of string concatenation? 

For example $"{settings.BaseUrl}/swagger..."

au5ton: On it. I didn't think that C# had syntax like Python fstrings or js string literals.

tidusjar:
Excellent! Thanks a lot for this! I'll wait for the build to complete then merge it in 😊

Issue #3195, opened 2019-10-10
album cover urls being broken by http replace


kionay:
**Describe the bug**
Album covers don't properly load sometimes.

**To Reproduce**
1. go to search for music
2. ensure the checkbox for 'Album Search' is clicked
3. search for something like 'Sound of Silence'
4. see many album cover images not loading

**Expected behavior**
Album cover URLs should be retrieved and properly made into https URLs when appropriate

**Desktop (please complete the following information):**
 - OS: windows

**Ombi Version (please complete the following information):**
 - Version 3.0.4680
- Media Server Plex

**Additional context**
Ombi properly connected to Lidarr
No logs because ombi doesn't realize it's an error.
when doing a search for an album you can also open up your browser's development tools of choice and view the console to see errors for unknown URL schemes 
Urls from coverartarchive.org seems to sometimes return https urls, and http is being replaced with https turning the url into https**s**://...

There are a few places with `.Replace("http", "https");` that could be the culprit, but it seems like lines 169 and 190 in src/Ombi.Core/Engine/MusicSearchEngine.cs are the most likely.
I'm not in a place to make code changes, nor have the ability to test them to see if that really is the root cause.

Perhaps something like this would help
```cs
vm.Cover = new UriBuilder(vm.Cover){Scheme=Uri.UriSchemeHttps}.ToString();
```

jakkalas:
I'd like to have a look at this one

tidusjar:
This has been fixed in the develop branch! 🎉 🎈

PR #3199, opened 2019-10-12
Fixes issue #3195

msdeibel:
The new string extension method ToHttpsUrl ensures that URLs starting with "https" are no longer turned into "httpss"
The commit also replaces all occurances of the error prone .Replace("http", "https") in the whole solution.


tidusjar:
Looks great! Thank you!

Issue #3153, opened 2019-09-13
Wrong notification "approved" when request is refused


yoriiis:
**Describe the bug**
The notification send on Slack when a request is refused is incorrect. 
```
Hello! Your request for {Title} on {ApplicationName}! This is now available! :)
```

**To Reproduce**
1. Open Ombi with a user which need request approval
2. Find a movie
3. Request the movie
4. Open Ombi with the admin account
5. Click on the button `Deny` to refused the request
6. Open Slack to see the notification send after the deny request
7. Notification `Request Available` is show

**Expected behavior**
The Slack notification send after the deny request should be `Request Declined`, with the default text:
```
Hello! Your request for {Title} has been declined, Sorry!
```
**Desktop (please complete the following information):**
 - OS: Ubuntu server

**Ombi Version (please complete the following information):**
 - Version: 3.0.4680
 - Media Server: Emby

**Additional context**
Slack notification enabled with webhook
Two accounts on Ombi : Admin and one simple user

tidusjar:
This has been fixed in the develop branch! 🎉 🎈

yoriiis:
Great news, thanks ! 👍

PR #3201, opened 2019-10-12
Fixes issue #3153

Tim-Trott:
It looks like the Slack notification agent was using the wrong notification type in the request declined method.


Tim-Trott:
#3153 

Just linking to the issue

tidusjar:
Excellent! Thanks a lot!

Issue #3213, opened 2019-10-18
Allow hiding of certain titles


gtbuchanan:
I'd like a way to hide certain titles. For instance, I don't want stuff like this showing up by default on the main page:

![image](https://user-images.githubusercontent.com/715687/67097791-7a365600-f180-11e9-90f9-be1a62b24002.png)

I'm not entirely sure why it's showing anyway since it was released in 2015 and doesn't seem like something that would fit the "Popular Movies" filter.

A "Hide" button on each title would be nice to avoid situations like this in the future. Some sort of admin-side TMDB "keyword" filter would probably be helpful too since this particular movie is labelled "erotic movie".

Running v3.0.4817 in `linuxserver/docker-ombi` docker image

PR #3215, opened 2019-10-19
Add adult movie filtering

gtbuchanan:
Adds a new settings section for The Movie Database to configure adult movie filtering.

image

image

Adult movies are already hidden by default, but there are some lewd 18+ movies that are not marked as adult movies that I have seen show up in suggestions (see #3213). I had to change all the "movie suggestion" methods to use the `discover/movie` API since the fixed movie APIs do not have the necessary filters. I configured the parameters so the results match exactly what was displayed before. Unfortunately, the movie search endpoint doesn't allow filtering by "keyword" so the keyword filter in Ombi only applies to movie suggestions.

I also cleaned up query parameter handling a bit since I noticed some duplication.


tidusjar:
Nice work! I have a quick look over the code and all looks ok. I'll do some testing tonight. 

Thanks a lot! 👌

src/Ombi.TheMovieDbApi/TheMovieDbApi.cs
```diff
@@ -99,65 +105,95 @@ public async Task<List<MovieSearchResult>> SimilarMovies(int movieId, string lan
...
         }
 
-        public async Task<List<MovieSearchResult>> PopularMovies(string langageCode)
+        public async Task<List<MovieSearchResult>> PopularMovies(string langCode)
         {
-            var request = new Request($"movie/popular", BaseUri, HttpMethod.Get);
-            request.FullUri = request.FullUri.AddQueryParameter("api_key", ApiToken);
-            request.FullUri = request.FullUri.AddQueryParameter("language", langageCode);
+            var request = new Request($"discover/movie", BaseUri, HttpMethod.Get);
+            request.AddQueryString("api_key", ApiToken);
```
tidusjar: Can we add
```
request.AddQueryString("sort_by", "popularity.desc");
```

src/Ombi.TheMovieDbApi/TheMovieDbApi.cs
```diff
@@ -99,65 +105,95 @@ public async Task<List<MovieSearchResult>> SimilarMovies(int movieId, string lan
...
-        public async Task<List<MovieSearchResult>> Upcoming(string langageCode)
+        public async Task<List<MovieSearchResult>> Upcoming(string langCode)
         {
-            var request = new Request($"movie/upcoming", BaseUri, HttpMethod.Get);
-            request.FullUri = request.FullUri.AddQueryParameter("api_key", ApiToken);
-            request.FullUri = request.FullUri.AddQueryParameter("language", langageCode);
+            var request = new Request($"discover/movie", BaseUri, HttpMethod.Get);
+            request.AddQueryString("api_key", ApiToken);
+            request.AddQueryString("language", langCode);
+            request.AddQueryString("with_release_type", "2|3");
```
tidusjar: Why do we need to specify the release type?

src/Ombi.TheMovieDbApi/TheMovieDbApi.cs
```diff
@@ -99,65 +105,95 @@ public async Task<List<MovieSearchResult>> SimilarMovies(int movieId, string lan
...
+            var startDate = DateTime.Today.AddDays(7);
+            request.AddQueryString("release_date.gte", startDate.ToString("yyyy-MM-dd"));
+            request.AddQueryString("release_date.lte", startDate.AddDays(17).ToString("yyyy-MM-dd"));
```
tidusjar: Can you please explain what is going on with the dates?

src/Ombi.TheMovieDbApi/TheMovieDbApi.cs
```diff
@@ -99,65 +105,95 @@ public async Task<List<MovieSearchResult>> SimilarMovies(int movieId, string lan
...
         {
-            var request = new Request($"movie/now_playing", BaseUri, HttpMethod.Get);
-            request.FullUri = request.FullUri.AddQueryParameter("api_key", ApiToken);
-            request.FullUri = request.FullUri.AddQueryParameter("language", langageCode);
+            var request = new Request($"discover/movie", BaseUri, HttpMethod.Get);
+            request.AddQueryString("api_key", ApiToken);
+            request.AddQueryString("language", langCode);
+            request.AddQueryString("with_release_type", "2|3");
+            var today = DateTime.Today;
+            request.AddQueryString("release_date.gte", today.AddDays(-42).ToString("yyyy-MM-dd"));
```
tidusjar: Also why these dates too?

src/Ombi/ClientApp/app/settings/themoviedb/themoviedb.component.html
```diff
@@ -0,0 +1,29 @@
...
+                <label for="showAdultMovies" tooltipPosition="top" pTooltip="Include adult movies (pornography) in results">Show Adult Movies</label>
+            </div>
+        </div>
+
+        <div class="form-group">
+            <label for="excludedKeywordIds" class="control-label" pTooltip="Prevent movies with certain keywords from being suggested. May require a restart to take effect.">
+                Excluded Keyword IDs for Movie Suggestions
+            </label>
+            <div>
+                <input type="text" [(ngModel)]="settings.excludedKeywordIds" class="form-control form-control-custom " id="excludedKeywordIds"
```
tidusjar: Might be worth using some sort of tag box instead here, for example:
https://www.primefaces.org/primeng/#/chips

tidusjar: We could also use the Keyword search API so the user can easily lookup keywords instead of attempting to find the ID's?
https://developers.themoviedb.org/3/search/search-keywords

What do you think?

src/Ombi.TheMovieDbApi/TheMovieDbApi.cs
gtbuchanan: @tidusjar Yes, I can add. I was torn between including it because it is more explicit or excluding it because it is the [default](https://developers.themoviedb.org/3/discover/movie-discover) and fewer bits across the wire.

src/Ombi.TheMovieDbApi/TheMovieDbApi.cs
gtbuchanan: @tidusjar To achieve parity with `movies/upcoming` since [the docs say](https://developers.themoviedb.org/3/movies/get-upcoming) that endpoint uses these release types.

src/Ombi.TheMovieDbApi/TheMovieDbApi.cs
gtbuchanan: @tidusjar I did the best I could to emulate `movie/upcoming` here, but the date logic for that endpoint is unclear. `movie/upcoming` sends back an object like this, but doesn't specify how it arrived at the values:

```json
"dates": {
    "maximum": "2016-09-22",
    "minimum": "2016-09-01"
}
```

Therefore, I had to assume the minimum and maximum were a certain offset from today. I compared the results I got with the results from `movie/upcoming` and they were the same for now. I will add comments to make this more clear.

src/Ombi.TheMovieDbApi/TheMovieDbApi.cs
gtbuchanan: @tidusjar Same reason as `movie/upcoming` for `movie/now_playing`. The endpoint returns an arbitrary date range without specifying in the docs what logic it is using.

src/Ombi/ClientApp/app/settings/themoviedb/themoviedb.component.html
gtbuchanan: @tidusjar I generally agree, as that was my original thought. I was mostly just trying to get out a working filter so I could get the specified movie off my home page since my parents and other family use Ombi. However, I also wasn't sure about how to handle the [different separators](https://developers.themoviedb.org/3/discover/movie-discover) in this case since you can use `,` (`AND`) or `|` (`OR`). I tested with both and they seem to provide the same results for the `without_keyword` filter, but I wasn't really confident enough to just pick using one over the other. I'm okay with just picking one like I did in the `placeholder` if that's what you would prefer.

gtbuchanan: @tidusjar A couple questions:
1. [ngx-chips](https://github.com/Gbuomprisco/ngx-chips) supports [autocomplete within their control](https://angular-mfppay.stackblitz.io/) ([unlike primeng](https://github.com/primefaces/primeng/issues/5551)). With something like that, a separate search control would be unnecessary. I know Ombi is already referencing `primeng`, but would you be opposed to me adding a reference to `ngx-chips` as well for this?
2. TMDB doesn't provide an endpoint to retrieve multiple keywords at a time by ID so we would likely still want to show the keyword ID in each chip to avoid having to send `n` queries on load. I was thinking the autocomplete could search by the keyword (`/search/keyword`) but clicking on an existing chip would show the name (`/keyword`) in a popover or something. How does that sound?

tidusjar: 1. Yeah that's fine, I only referenced primeng because I knew they had an existing control. But if you have found a better one then I'm all for it. 
2. That sounds like a good idea.

gtbuchanan:
New usability notes:

image

gtbuchanan:
@tidusjar I've completed the requested changes, though, the GitHub build failed on `SetVersionInfo` for some reason.

tidusjar:
Looking good! Nice work! 

I'll do some testing soon

tidusjar:
Nice! All merged in! Thanks for this.

If you are looking to get involved a bit more let me know! It would be awesome to have you on board

PR #3216, opened 2019-10-20
Fix search bar overlap on mobile

gtbuchanan:
Fixes the "Suggestions" and "Refine" buttons overlapping the search bar on mobile devices (specifically phones) in portrait mode. The alternative view is only present when the width is 450px or less.

Before            |  After
:-------------------------:|:-------------------------:
 | 

Ignore the "TV Shows" tab missing from the "Before" as I have it hidden on my prod server. Also ignore the icons missing from the "After" as I'm not sure why they don't show up in my local environment (any pointers on fixing this?).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## (unreleased)
+## v3.0.4817  (2019-10-15)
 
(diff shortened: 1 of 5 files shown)


tidusjar:
Thanks!

PR #3291, opened 2019-11-29
Add custom webhook notifications

Namaneo:
Adds the ability to configure a simple webhook for notification sending. Instead of sending a formated message, the whole dictionary of notification properties is sent as a JSON.

You can test it by running a simple webhost and by configuring one of its endpoints in Ombi. Test payload will by empty, so this may be better to trigger a real notification instead. 🙂

I tried to be as compliant as possible with your code-base, don't hesitate to tell me if some pieces of code don't fit you !


tidusjar:
Wow nice job! i'll take a look at the code and do some testing in the next day or two!

tidusjar:
There are some linting errors on the build:

```
ERROR: C:/projects/requestplex/src/Ombi/ClientApp/app/services/applications/tester.service.ts:15:5 - Named imports must be alphabetized.
ERROR: C:/projects/requestplex/src/Ombi/ClientApp/app/services/settings.service.ts:18:5 - Named imports must be alphabetized.
ERROR: C:/projects/requestplex/src/Ombi/ClientApp/app/settings/notifications/webhook.component.ts:4:10 - Named imports must be alphabetized.
ERROR: C:/projects/requestplex/src/Ombi/ClientApp/app/settings/settings.module.ts:33:1 - Import sources within a group must be alphabetized.
```

tidusjar:
I just had a quick skim through the code and it looks all good!

Just a few points:
1. The access token should be optional, and we shouldn't apply the header if the value is `null` (A lot of webhook providers provide the key/access token as part of the uri string
1. It would probably be quite useful for us to include the Notification Type as part of the payload e.g. NewRequest, RequestApproved
1. Since we have the notification types, it would be good if the user could specify which types it wanted to consume
1. We could add the ability to push to multiple webhooks.

Let me know what you think of the above functionality


Thanks again!

Namaneo:
I fixed the linting issues, sorry for that ! For what you stated:

1. Access token is already optional in the UI, but indeed I updated the API to prevent adding it if not configured.
2. Shame on me, I was so sure it was already on the variables dictionary I didn't even check... I added it ! Related to the uri string, I added a "token" parameter, but this name may change depending on the webhook provider you use, right? And in that case, it may even be better to be able to configure header and uri parameter names?
3. I assumed that this webhook configuration was for custom-developed ones, and therefor it would have been the target API's responsibility to ignore messages or not. Do you agree?
4. Indeed we could, and this would be nice ! But for now I'm slowly discovering the project and its architecture, I would be glad to help with that but maybe later?

Let me know if something else needs to be changed. 🙂

tidusjar:
Thanks for the changes! 

I think I didn't quite explain correctly regarding having the token in the Uri. So for example they could just put the token as part of the webhook URL, since you are correct, the query name could change on the consumer. 

So I don't think we need to do anything here regarding that, other than maybe a note on the UI stating the access token will be passed as a header called "AccessToken"

Regarding the other points, yes they could be done at a future date!

Namaneo:
Added a tooltip for the token configuration on the settings view 🙂

tidusjar:
Awesome. I'll do a bit of testing tonight!

src/Ombi.Api.Webhook/WebhookApi.cs
```diff
@@ -0,0 +1,40 @@
...
+        public async Task PushAsync(string baseUrl, string accessToken, IDictionary<string, string> parameters)
+        {
+            var request = new Request("/", baseUrl, HttpMethod.Post);
+
+            if (!string.IsNullOrWhiteSpace(accessToken))
+            {
+                request.AddHeader("Access-Token", accessToken);
+            }
+
+            var body = parameters.ToDictionary(
```
tidusjar: `parameters` can be null. e.g. when testing

Namaneo:
Improved test notification to use a more accurate payload: parameters should not be null anymore !

tidusjar:
Nice job dude! 

Thanks for the effort you have put into this!

PR #3297, opened 2019-12-05
Add ability to approve or deny a request directly from a Telegram chat

Namaneo:
Alright, this one is more like a proof of concept than the end feature. In fact, I wanted to have your advices regarding the organization of this code.

Basically, all is in the title: this adds the ability to approve or deny a request directly from the configured Telegram chat. It makes use of the [Telegram.Bot](https://github.com/TelegramBots/Telegram.Bot) library to help sending and receiving more complex messages. As long a the Telegram settings are properly configured, for every `NewRequest` notification, two buttons and the poster image are joined to the message, and you can one-click answer.

However, here are few questions that does not have answers yet:
* Right now, Telegram callbacks configuration is done in `Startup.cs` but this is probably not a good place. Knowing that `Ombi.Telegram.Api`can't have a dependency to the `IMovieRequestEngine` and `ITvRequestEngine`, where could this piece of code be moved? (moving it might as well remove some complexity in `TelegramRequestAnswer` implementation)
* For the most related to the previous question, there's some places where I hard-coded strings because of the lack of dependencies. Is it an issue for you, as long as it remains small specific cases?
* Do you think it would be a good idea to go even further with this kind of feature? By saying that, I mean allowing non-admin users to search and request items through the bot directly as well. I would love to help developing this kind of stuff, but I'm honestly not sure it belongs directly to Ombi.

I should certainly add some comments in the code to help describing what's going on, but since I'm not sure how it will change depending on you suggestions, I will do that at the end. :)

### DIFF
--- a/src/Ombi.Api.Telegram/ITelegramApi.cs
+++ b/src/Ombi.Api.Telegram/ITelegramApi.cs
@@ -1,9 +1,10 @@
-﻿using System.Threading.Tasks;
+﻿using System.Collections.Generic;
+using System.Threading.Tasks;
 
 namespace Ombi.Api.Telegram
 {
     public interface ITelegramApi
     {
-        Task Send(string message, string botApi, string chatId, string parseMode);
+        Task Send(string message, string botApi, string chatId, string parseMode, IDictionary<string, string> data);
     }
 }
--- a/src/Ombi.Api.Telegram/ITelegramRequestAnswer.cs
+++ b/src/Ombi.Api.Telegram/ITelegramRequestAnswer.cs
@@ -1,0 +1,12 @@
+﻿using System.Threading.Tasks;
+
+namespace Ombi.Api.Telegram
+{
+    public interface ITelegramRequestAnswer
+    {
+        Task<string> ApproveMovie(int id);
+        Task<string> DenyMovie(int id);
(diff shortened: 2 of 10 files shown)


tidusjar:
Hi! 

Thanks for this contribution again!

Now I've been thinking about this and I'm not sure this is something that I want to include inside Ombi.

I think this sort of functionality would be excellent as a separate application, and there are a few that do this e.g. [Mellow](https://github.com/v0idp/Mellow)

Now please keep in mind that I really appreciate the time you have put into this, but since I have to support it going forward, I'm not going to accept this PR into the application at this time.

I suggest if you get any other ideas maybe contact me first? Either my email or just jump on Discord :)

Namaneo:
Don't worry about the time I spent on it, I mostly took this code from a personal project. This PR was more like an experiment I've made on Ombi, and that I wanted to show you. No regrets here, this was interesting !

I fully uderstand your decision, this was kind of expected. I will continue to work on this on my separate project, as you suggested. :)

Btw, thanks for pointing out Mellow, didn't know about it, looks great !

tidusjar:
Hey but if you want to help out with the project, I'm always open for ideas and stuff! 

I'd be happy for you to contribute when you have time! 😃

PR #3313, opened 2019-12-17
Use URL redirection instead of popup for Plex OAuth

Namaneo:
IMO having an OAuth mechanism which redirects to and from Plex sign-in view is easier to use (especially on mobile browsers). As far as I've tested, those changes works great on my side. What do you think of it?  🙂

### DIFF
--- a/src/Ombi/ClientApp/app/login/login.component.ts
+++ b/src/Ombi/ClientApp/app/login/login.component.ts
@@ -79,7 +79,7 @@ export class LoginComponent implements OnDestroy, OnInit {
             this.router.navigate(["search"]);
         }
 
-        let pinId = this.route.snapshot.queryParams["pinId"];
+        const pinId = this.route.snapshot.queryParams.pinId;
         if (pinId) {
             this.notify.info("Authenticating", "Loading... Please Wait");
             this.getPinResult(pinId);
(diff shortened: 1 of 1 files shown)


tidusjar:
Haha, funny you should add this, it's been on my list of TODO's and it was originally implemented like this!

I'll do some testing, thanks for the contributions again!

JonnyWong16:
Redirection will not work in an iframe due to Plex's `X-Frame-Options` header (e.g. when Ombi is framed in Organizr).

tidusjar:
That is a very good point.

It might be worth letting the user decide on how they want to configure this and support both methods.

Namaneo:
Indeed, I could reproduce easily. I tried few things, and ended by redirecting the whole top frame instead of the embedded one. This works, though this might not be the best solution: Plex redirects to Ombi instead of Organizr. 

One solution would be the have a middleware empty view between Plex OAuth view and Ombi login view: all this view would do is to store the `pinId` inside a cookie and redirects to Organizr. Will try that. :)

@tidusjar Are you fine with cookie usage? In this specific case, it will only stay during the authentication process (as there is not need of it after the `access_token` is retrieved)

bdr99:
I noticed that this PR got closed by stalebot. Is this feature still being considered? I'd really like to see this functionality in Ombi as it really enhances the user experience!

Issue #3253, opened 2019-11-13
Movie search doesn't work after upgrade to v3.0.4880


paulfantom:
**Describe the bug**
Section `Movies` in Search tab doesn't show any movies and don't return any results after trying to search. There is no such problem in `TV Shows` tab.

**To Reproduce**
1. Go to 'Search'
2. Click on 'Movies' (already at this point there is no proposed movies)
3. Enter any movie title and press Enter or click on looking glass icon
4. Nothing happens

**Expected behavior**
Movie propositions are displayed and search works.

**Screenshots**
![Screenshot_20191113_110352](https://user-images.githubusercontent.com/3531758/68753571-91445880-0605-11ea-9769-15be6af3a3d7.png)

**Logs (Logs directory where Ombi is located)**
I think this one is relevant:
```
fail: Ombi.Schedule.Jobs.Plex.PlexContentSync[2008]
      Exception when adding new Movie <NAME_REDACTED>
Microsoft.Data.Sqlite.SqliteException (0x80004005): SQLite Error 1: 'no such column: x.RequestId'.
   at Microsoft.Data.Sqlite.SqliteException.ThrowExceptionForRC(Int32 rc, sqlite3 db)
   at Microsoft.Data.Sqlite.SqliteCommand.PrepareAndEnumerateStatements(Stopwatch timer)+MoveNext()
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReader(CommandBehavior behavior)
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReaderAsync(CommandBehavior behavior, CancellationToken cancellationToken)
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteDbDataReaderAsync(CommandBehavior behavior, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Storage.Internal.RelationalCommand.ExecuteAsync(IRelationalConnection connection, DbCommandMethod executeMethod, IReadOnlyDictionary`2 parameterValues, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Query.Internal.AsyncQueryingEnumerable`1.AsyncEnumerator.BufferlessMoveNext(DbContext _, Boolean buffer, CancellationToken cancellationToken)
   at Microsoft.EntityFrameworkCore.Query.Internal.AsyncQueryingEnumerable`1.AsyncEnumerator.MoveNext(CancellationToken cancellationToken)
[...]

PR #3314, opened 2019-12-18
Undo this recent addition as it broke movie search

ryanwkirby:
Hopefully fix bug #3253


tidusjar:
This will not fix the issue. 

The issue is to do with the EF migrations and it's not solvable now due to the change of the migration files to match the new database formats.

PR #3344, opened 2020-01-11
Add RequestId variable in notification's data

Namaneo:
Just adding the `RequestId` property into the notification's data. I believe this would essentially help for the Webhook kind. For instance, in my case, this will allow me to easily keep track of a request from a Telegram bot.

### DIFF
--- a/src/Ombi.Core/Rule/Rules/Search/ExistingRule.cs
+++ b/src/Ombi.Core/Rule/Rules/Search/ExistingRule.cs
@@ -63,7 +63,6 @@ namespace Ombi.Core.Rule.Rules.Search
                     request.Requested = true;
                     request.Approved = tvRequests.ChildRequests.Any(x => x.Approved);
                     request.Denied = tvRequests.ChildRequests.Any(x => x.Denied ?? false);
-                    request.DeniedReason = tvRequests.ChildRequests.FirstOrDefault(x => x.DeniedReason != null)?.DeniedReason;
 
                     // Let's modify the seasonsrequested to reflect what we have requested...
                     foreach (var season in request.SeasonRequests)
--- a/src/Ombi.Core/Rule/Rules/Search/ExistingRule.cs
+++ b/src/Ombi.Core/Rule/Rules/Search/ExistingRule.cs
@@ -34,6 +34,8 @@ namespace Ombi.Core.Rule.Rules.Search
                     obj.Requested = true;
                     obj.RequestId = movieRequests.Id;
                     obj.Approved = movieRequests.Approved;
+                    obj.Denied = movieRequests.Denied ?? false;
+                    obj.DeniedReason = movieRequests.DeniedReason;
                     obj.Available = movieRequests.Available;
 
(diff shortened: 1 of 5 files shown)


Namaneo:
Added two commits:
* One to execute search rules while searching a TV show (for instance using the `Search/tv/{searchTerm}` endpoint). Was there a reason why they weren't executed?
* Added `Denied` and `DeniedReason` on searches. Again, helps for keeping track of a request while querying the API. Since there was already `Requested`, `Approved` and `Available`, I think this is relevant?

Side question: in the future, for this kind of changes, do you prefer one PR with small commits, or multiple PR, one for each? 🙂

tidusjar:
Hey man!

Thanks for the contributions!

I'll take a look at them sometime soon, been super busy with IRL stuff and work.

But to answer your question, personally I find multiple PR's easier, since it's clear what that PR is trying to accomplish and easier to merge and test. But no need to change this one :)

Thanks!

src/Ombi.Core/Rule/Rules/Search/ExistingRule.cs
```diff
@@ -60,6 +62,8 @@ public async Task<RuleResult> Execute(SearchViewModel obj)
 
                     request.Requested = true;
                     request.Approved = tvRequests.ChildRequests.Any(x => x.Approved);
+                    request.Denied = tvRequests.ChildRequests.Any(x => x.Denied ?? false);
```
tidusjar: I'm not sure how I feel about this. Since we can have many child requests for one show. If one of them is denied it will seem like the whole show has been denied.

Namaneo: I agree, I took the line above as an example. I believed that if `Approved` was initialized like that, it would be OK for `Denied` as well. 

One solution would be to have the same kind of mechanism as `request.FullyAvailable` and `request.PartlyAvailable`, for `Approved` and `Denied`. Not sure about the complexity, since it would probably affect the UI. What do you think?

tidusjar: Good point, we don't show the approved as part of the UI. But good point.

I think Denied in this case is probably fine, but can we remove the Denied Reason? since that could vary if there are multiple denies, if someone wants to see the reason for denial then they can view the child request.

Namaneo: Either that, or we format a comma-separated list of all reasons: this might get messy in some cases, but at least a relevant value is returned? 

Unless you say otherwise, I'll remove this line. Waiting for your final word to push. 🙂

tidusjar: Yeah I think that would get quite messy if we had 10 denied requests with 50 chars each for example! :)

Namaneo:
All done on my side, is there anything else you want me to update?

tidusjar:
All looks good! Thanks a lot!

PR #3404, opened 2020-02-22
Refuse a suggestion

Alandil35:
Hello,
I think it would be interesting to be able to reject a suggestion as proposed by Radarr. Indeed, on each film proposal, we can hide / refuse the proposal. What do you think about this ?


tidusjar:
Hello, please open up a issue and not a pull request

PR #3481, opened 2020-04-09
Fix IMDB links for TV shows not working

nirmalspeed:
Links for TV shows were just relative paths to their imdb ID. 
ex: `href="tt0944947"` for Game of Thrones


tidusjar:
Thanks!

PR #3503, opened 2020-04-19
Goldenpipes tv child

goldenpipes:
Moved the Requested by user from the child section to the main section on the TV request page. it just visually cleans things up imho.

hope full i did this right!


tidusjar:
So the reason requested by was on the child and not the main TV show is because User A can request season 1 and User B can request season two, so there is not a single user for TV.

goldenpipes:
ah I never thought of that!  makes perfect sense though. Kill this.

PR #3539, opened 2020-05-10
Feature/v4

twanariens:
A big CSS overhaul.


src/Ombi/ClientApp/src/app/settings/emby/emby.component copy.html
```diff
@@ -0,0 +1,110 @@
+﻿
```
tidusjar, line 1: Is this file needed?

twanariens: The copy is there as a failsafe. I made changes to the emby.component.html to make it use mat-tab-group. But I don't have Emby running so I am unable to check if it is working as intended.

PR #3598, opened 2020-06-04
Replace smartquotes in meta info

jonbloom:
With the smartquotes, only the first word of the custom app gets loaded in inside apps that use those tags for previews.

For example, my Ombi instance has the name set as "Bloom Plex Requests" but when sent via iMessage, the preview only shows "Bloom as the title.


jonbloom:
Hey @tidusjar 

I didn't see a way to notify you in any way beside a comment. Can you take a look at this PR?

Thanks!

tidusjar:
Hey! Thanks! :D

Issue #3587, opened 2020-05-28
Consider moving release tags to the same repo as the code


frebib:
**Describe the bug**
Today I came to build Ombi for our Docker image to find that the beta v4 "release" tags are in a separate repo (I can't work out why) and unhelpfully the git tags are on a bogus commit. The whole point of generating releases alongside the code is so there is a relationship between source and binary. There is no way of programatically mapping the tree from which the release was generated, and the tag attached to the release. I can't even use the github source tar because it is for the other repo.

Please consider putting the source and the release tags in the same repo; either reinstating the tags in this repo, or duplicating the source tree into the release repo and retaining the tags there. As it currently stands, I have no choice but to either build HEAD (probably fine, but there is little point of reference for breakages, and no version numbers) or use your precompiled source (which I'd rather not do, I want it to work on Alpine)

**To Reproduce**
1. Nope
2. Not relevant
3. Why do you enforce this stupid template?

**Expected behavior**
To not have my valid issues closed by a fucking bot

**Screenshots**
No

**Logs (Logs directory where Ombi is located)**
...

**Desktop (please complete the following information):**
Not every bug report is user error

**Ombi Version (please complete the following information):**
 - If there were tags I would know

**Additional context**
Your bot is pointless

tidusjar:
Wow, what's with all the attitude in this issue.

Regards to the separate repo, this is not a new concept I have just "made" up. For example, the Emby Team do a similar thing: https://github.com/MediaBrowser/Emby.Releases/releases

If you are that desperate to link up the commits with the build, you can just check the build pipelines. https://dev.azure.com/tidusjar/Ombi/_build?definitionId=18&_a=summary

So thanks for you input, but I'm not going to change this just because you don't like it.

Also next time, please try and ask questions a bit nicer, this is an open source project, not a paid for service. Keep that in mind.

frebib:
Sorry about the frustration coming out, I was pretty pissed when I wrote this- not exactly good form. By apologies. I do appreciate your work on this, it's very useful

Emby isn't exactly a good model to live by given their more recent history, IMO.

Perhaps if you don't want to change the releases repo, could you instead put the release tags on this repo too aswell?
Looking at the CI, I can't see where the release tags come from? Are they automatically generated?

frebib:
An even easier method would be to upload a source tarball in the releases repo alongside the built binaries. Github provides source tarballs, but they are for the local repo, so in this case would contain the README file for the releases repo.

[...]

PR #3632, opened 2020-06-20
Create tidusjar/Ombi pre-release tags

frebib:
isDraft creates unpublished tags, which are not publicly visible. This
change publishes the tags as pre-release github releases with tags.

Fixes tidusjar/Ombi#3587

Signed-off-by: Joe Groocock <me@frebib.net>

### DIFF
--- a/.azuredevops/pipelines/publish-job.yml
+++ b/.azuredevops/pipelines/publish-job.yml
@@ -70,7 +70,7 @@ stages:
         target: '$(Build.SourceVersion)'
         tagSource: 'userSpecifiedTag'
         tag: '$(gitTag)'
-        isDraft: true
+        isPreRelease: true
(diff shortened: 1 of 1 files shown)


tidusjar:
I don't currently want to publish the builds to two places, it's currently going to https://github.com/tidusjar/ombi.releases/

I will soon switch over the binaries to this repo, but not yet, need to work with a few teams e.g. LSIO to switch their dockers over

frebib:
As-is this task in the Azure pipeline does nothing.
In #3587 I requested git tags for "release" commits and that is what you added.
Can we make this add only git tags and not a github release?

tidusjar:
So my plan is just to have the pre-releases in this repo, this will happen soon.

frebib:
That's even better, thanks!

Issue #3691, opened 2020-07-28
Integration with Jellyfin Externally Facing Hostname


gwenvador:
**Describe the bug**
When clicking on the movie item to connect to Jellyfin it does not open the correct link. 

**To Reproduce**
1. Go to any Movie
2. Click on the Jellyfin Link
3. See error

**Expected behavior**
Access to the movie in Jellyfin at the right address. The current link added to Externally facing Hostname is #!/itemdetails.html?id=1 it should be #!/details?id=1.

I am not sure if it is historical or the way that emby is configured but we should be able to configure the full external facing link as jellyfin is not configured with itemdetails. 

**Ombi Version (please complete the following information):**
 - Version 4.0.414.0
- Media Server Jellfyin 10.6.0

CatBox360:
I myself am also having this issue with both the latest v3 and v4.

In addition to the above, in between the index.html and the #! there is / that should not be there for Jellyfin

The "View on Emby" button provides: https://my.server/web/index.html/#!/itemdetails.html?id=12345
Jellyfin provides this link when navigating: https://my.server/web/index.html#!/details?id=12345

Everything else seems to be working as expected.

gwenvador:
Good point.
I just put a pull request for the changes. Hopefully this works.

juligreen:
This is sadly not fixed, could it please be reopened? :)

tidusjar:
Please reopen a new bug. But I'm pretty sure this is fixed in v4

PR #3695, opened 2020-07-29
Update EmbyHelper.cs

gwenvador:
The link provided for connecting to Jellyfin is not working. I updated the path to reflect the change and kept the link for not Jellyfin users. Issue #3691


tidusjar:
We need to update the unit tests

PR #3746, opened 2020-09-03
settings textbox width fix

goldenpipes:
this makes the text entry boxes in the settings pages fill their element instead of being so narrow.

image

vs OLD

image

IM LEARNING!


tidusjar:
Thanks!

PR #3747, opened 2020-09-04
Arr's settings pages

goldenpipes:
made the settings pages of all the *arrs match eachother.

im kinda getting the hang of how the pages are designed!


goldenpipes:
im re-doing this again.

PR #3783, opened 2020-10-03
Added logging of failed login for Fail2Ban

quietsy:
Added logging of failed login for Fail2Ban, supports reverse proxy and direct connection, tested directly and using NGINX.

Fail2Ban jail for testing:

[INCLUDES]
before = common.conf

[Definition]
failregex = ^.*Failed login attempt by IP: <ADDR>.*$
ignoreregex =

### DIFF
--- a/src/Ombi/Controllers/V1/TokenController.cs
+++ b/src/Ombi/Controllers/V1/TokenController.cs
@@ -7,6 +7,7 @@ using System.Text;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.Options;
+using Microsoft.Extensions.Logging;
 using Microsoft.IdentityModel.Tokens;
 using Ombi.Core.Authentication;
 using Ombi.Helpers;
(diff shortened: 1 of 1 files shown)


src/Ombi/Controllers/V1/TokenController.cs
```diff
@@ -80,6 +84,7 @@ public async Task<IActionResult> GetToken([FromBody] UserAuthModel model)
                 var url = await _plexOAuthManager.GetOAuthUrl(model.PlexTvPin.code, websiteAddress);
                 if (url == null)
                 {
+                    _log.LogWarning(string.Format("Failed login attempt by IP: {0}", GetRequestIP()));
```
tidusjar, line 87: This is not really a failure due to the user here

tidusjar:
Nice work, all good except one small comment!

tidusjar (changes requested):
Just one small comment! Nice work

src/Ombi/Controllers/V1/TokenController.cs
quietsy: You're right, thanks!

quietsy:
Thanks!
You were right, I added another commit and tested it just in case.

tidusjar (approved):

tidusjar:
Sweet! Nice work!

Issue #3773, opened 2020-09-24
Failed to load API definition


Emtech2:
**Describe the bug**
When I attempt to go to the Swagger documentation I get a 500 internal server error. This also happens if I go to /swagger/v1/swagger.json
This is the same error as #3651 which was closed due to no responses

**To Reproduce**
1. Go to ombi.example.com/swagger
2. See the error
or
1. Go to ombi.example.com/swagger/v1/swagger.json
2. See error in JSON

**Expected behavior**
To see the Swagger logs

**Screenshots**
https://i.imgur.com/RR6uoR8.png

**Logs (Logs directory where Ombi is located)**

```2020-09-24 04:00:09.469 -04:00 [Error] Something bad happened, ErrorMiddleware caught this
 System.NotSupportedException: Conflicting method/path combination "POST api/v1/Radarr/tags" for actions - Ombi.Controllers.V1.External.RadarrController.GetTags (Ombi),Ombi.Controllers.V1.External.RadarrController.GetTags (Ombi). Actions require a unique method/path combination for Swagger/OpenAPI 3.0. Use ConflictingActionsResolver as a workaround
    at Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.GenerateOperations(IEnumerable`1 apiDescriptions, SchemaRepository schemaRepository)
    at Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.GeneratePaths(IEnumerable`1 apiDescriptions, SchemaRepository schemaRepository)
    at Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.GetSwagger(String documentName, String host, String basePath)
    at Swashbuckle.AspNetCore.Swagger.SwaggerMiddleware.Invoke(HttpContext httpContext, ISwaggerProvider swaggerProvider)
    at Microsoft.AspNetCore.Authorization.AuthorizationMiddleware.Invoke(HttpContext context)
    at Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.Invoke(HttpContext context)
    at Ombi.ApiKeyMiddlewear.Invoke(HttpContext context) in /home/vsts/work/1/s/src/Ombi/Middleware/ApiKeyMiddlewear.cs:line 58
    at Ombi.ErrorHandlingMiddleware.Invoke(HttpContext context) in /home/vsts/work/1/s/src/Ombi/Middleware/ErrorHandlingMiddlewear.cs:line 30
```

[...]

PR #3798, opened 2020-10-07
Resolves endpoint conflict for swagger.

bernarden:
Fixes the issue #3773 introduced in feature/v4 branch. 
I wasn't able to reproduce #3768 with the code that is currently in master branch.


tidusjar:
Thanks!

Issue #3295, opened 2019-12-02
mixed content error


RobinMM:
**Describe the bug**
Firefox is giving mixed content error because http urls are used for music album images:
http://assets.fanart.tv/fanart/music/11ae9fbb-f3d7-4a47-936f-4c0a04d3b3b5/cdart/white-blood-cells-55fcd7d5b036b.png

**To Reproduce**
Request music album, your browser will then show mixed content error.

**Screenshots**
![afbeelding](https://user-images.githubusercontent.com/1236301/69956837-69a82800-1501-11ea-8c1a-f9e1b54c7c24.png)

**Desktop (please complete the following information):**
Running ombi Docker

**Ombi Version (please complete the following information):**
3.0.4892

Ghost519:
I'd like to try and take a shot at this issue, I think a https version of the domain exists since both http://assets.fanart.tv/fanart/music/11ae9fbb-f3d7-4a47-936f-4c0a04d3b3b5/cdart/white-blood-cells-55fcd7d5b036b.png and https://assets.fanart.tv/fanart/music/11ae9fbb-f3d7-4a47-936f-4c0a04d3b3b5/cdart/white-blood-cells-55fcd7d5b036b.png works, so the fix could be as easy as adding a 's' to the album pictures links

tidusjar:
Yeah sure go for it! We do this already for some Fanart images. There is already a helper method to also do this for us https://github.com/tidusjar/Ombi/blob/cd8a89800ae4c0a48027bd42d3661e303f3fd459/src/Ombi.Helpers/StringHelper.cs#L132

Ghost519:
never mind, it looks like I was way over my head and I couldn't do it, sorry for wasting your time

PR #3801, opened 2020-10-07
Replaces scheme to https for all urls returned to client side.

bernarden:
Fixes issue #3295.

### DIFF
--- a/src/Ombi.Helpers/StringHelper.cs
+++ b/src/Ombi.Helpers/StringHelper.cs
@@ -131,7 +131,7 @@ namespace Ombi.Helpers
 
         public static string ToHttpsUrl(this string currentUrl)
         {
-            return currentUrl.Replace("http://", "https://");
+            return currentUrl?.Replace("http://", "https://");
         }
     }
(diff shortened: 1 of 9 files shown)


src/Ombi.Helpers/StringHelper.cs
```diff
@@ -131,7 +131,7 @@ public static string StripCharacters(this string str, params char[] chars)
 
         public static string ToHttpsUrl(this string currentUrl)
         {
-            return currentUrl.Replace("http://", "https://");
+            return currentUrl?.Replace("http://", "https://");
```
tidusjar, line 134: Nice work! Looks good, 

Can we please add a unit test for this where the `currentlyUrl` is null?

tidusjar (changes requested):

src/Ombi.Helpers/StringHelper.cs
bernarden: Done. =)

tidusjar (approved):

Issue #2676, opened 2018-12-08
Request - TV episode air date on Newsletter


ePHDiSK:
#### Ombi build Version:

V 3.0.3988

#### Update Branch:

Master

#### Media Sever:

Plex

#### Problem Description:

It would be nice to have the episode(s) air date in the Newsletter next to the respective episode.  Currently it only shows the Show launch release year.  

#### Badly edited example: 

![image](https://user-images.githubusercontent.com/45718819/49689229-1ebd2580-fae4-11e8-8e33-ab7c24a3bf24.png)

#### Reproduction Steps:

Look at a generated newsletter with new TV episodes.

mariombaltazar:
Hi,
can i contribute with this issue?

tidusjar:
Yeah sure!!

peterszeles:
Hello,
i would like to contribute with this issue but I see @mariombaltazar took it on 28 august. Is it still actual or are you working on it?

tidusjar:
Nope, no one is working on this! 

If you have any issues or questions please feel free to reach out on discord.

peterszeles:
Ok, than I will try to fix this issue :) I have already cloned it to my pc. Are there any recommendations how to run this app?

tidusjar:
So the best way would be to run Visual Studio, you can use the Community Edition which is free.
And just set the Ombi.csproj as the startup project and Visual Studio will do all of the work for you.

I do have a more detailed guide on how to debug Ombi when the next major release comes out (due to slightly different frameworks).

AdamWorley:
I see this has been around for over year, mind if I try to fix it?

AdamWorley:
What would be the best way to format the air date?

I was thinking that for:
* single date - YYYY-MM-DD
* run of episodes start-end:
  * Same date - YYYY-MM-DD
  * Same month - YYYY-MM
  * Same year - YYYY-MM - MM
  * Multiple years - YYYY-MM - YYYY-MM

tidusjar:
Sure!

I would say for a run of episodes Same Month would be the best option

Abhi441:
I want to fix this bugs. Kindly Arange  zoom meating for give kt.

tidusjar:
@Abhi441 what?

sephrat:
This has been part of Ombi for quite some time now. Closing.

PR #3802, opened 2020-10-08
Pull request for issue #2676 to include episode aired date in newsletter

AdamWorley:
I have only implemented the aired date for newly added to plex as that's what was requested and I'm unfamiliar with emby.
issue #2676

### DIFF
--- a/src/Ombi.Api.Plex/Models/Metadata.cs
+++ b/src/Ombi.Api.Plex/Models/Metadata.cs
@@ -19,7 +19,7 @@ namespace Ombi.Api.Plex.Models
         public string banner { get; set; }
         public string theme { get; set; }
         //public string duration { get; set; }
-        public string originallyAvailableAt { get; set; }
+        //public string originallyAvailableAt { get; set; }
         public int leafCount { get; set; }
         public int viewedLeafCount { get; set; }
         public int childCount { get; set; }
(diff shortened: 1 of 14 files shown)


src/Ombi.Api.Plex/Models/Metadata.cs
```diff
@@ -19,7 +19,7 @@ public class Metadata
         public string banner { get; set; }
         public string theme { get; set; }
         //public string duration { get; set; }
-        //public string originallyAvailableAt { get; set; }
```
tidusjar, line 22: Can we please undo this change?

src/Ombi.Helpers/StringHelper.cs
```diff
@@ -77,6 +77,57 @@ public static int IntParseLinq(string stringIn)
             return -1;
         }
 
+        public static string BuildEpisodeList(IEnumerable<(int EpisodeNumber, DateTime? Aired)> orderedEpisodes)
```
tidusjar, line 80: Looks good, but don't feel like this belongs in the `StringHelper`? It's very specific to what's going on

tidusjar (changes requested):
We are now also adding the `Aired` property onto each of the entities, there needs to be data migrations created for SQLite and MySQL for this. Then they should also be populated when they are created (Not just in the RecentlyAddedEngine, e.g. the PlexContentSync). 

So far you have done a good job! Just it's not totally finished.

The other option is instead of storing the AirDate against the actual entity, in the NewsLetterJob we actually look up the media on TheMovieDb/TvMaze which should also contain that information, so we could easily leverage that

AdamWorley:
I have made the changes you (@tidusjar) have recommended and I have leveraged the Movie db for the season/episode info.

I opted to use the tv/season endpoint, instead of the episode endpoint to avoid spamming api requests.

PR #3805, opened 2020-10-10
WIP LDAP authentication

dpraul:
Never written any C# before this, so please bear with me here.

Included here is:
- A settings page for LDAP configuration
- LDAP authentication provided by [Novell.Directory.Ldap.NETStandard](https://github.com/dsbenghe/Novell.Directory.Ldap.NETStandard), based loosely on [`jellyfin-plugin-ldapauth`](https://github.com/jellyfin/jellyfin-plugin-ldapauth)
- An LDAP bulk importer
- The ability to create LDAP users on-the-fly at login (uses default roles from the user management settings)
    - Had to tweak the token creation a bit for this, but IMO it's a bit easier to follow now.

Potential TODO:
- [ ] Better documentation on settings page
- [ ] Don't return the bind password on the settings page
- [ ] Move ldap attribute mappings - ideally, to retrieve email, alias, etc. from directory
    - [ ] Updating the user attributes on-login when this happens (similar to in emby login)
- [ ] LDAP filter mappings to Ombi roles
    - [ ] Would also need update on-login

This is current running and functional in my setup - going to test it out for a bit and see if it needs any changes. Others are welcome to test, just know that YMMV. I pushed a docker image for this based on the Linuxserver image, which I'm currently running in my environment. It should be a drop-in replacement for `linuxserver/ombi:v4-preview`. You can find it at [`theactualdylan/ombi-tests:ldap-1`](https://hub.docker.com/r/theactualdylan/ombi-tests)

### DIFF
--- a/src/Ombi.Core/Authentication/LdapUserManager.cs
+++ b/src/Ombi.Core/Authentication/LdapUserManager.cs
@@ -1,0 +1,168 @@
+﻿using System;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Novell.Directory.Ldap;
+using Ombi.Core.Settings;
+using Ombi.Settings.Settings.Models;
+using Ombi.Store.Entities;
+
+namespace Ombi.Core.Authentication
+{
+    /// <summary>
+    /// Ldap Authentication Provider
+    /// </summary>
+    public class LdapUserManager : ILdapUserManager
+    {
+        public LdapUserManager(ILogger<LdapUserManager> logger, ISettingsService<LdapSettings> ldapSettings)
+        {
+            _ldapSettingsService = ldapSettings;
+            _logger = logger;
+        }
+
+        private readonly ISettingsService<LdapSettings> _ldapSettingsService;
+        private readonly ILogger<LdapUserManager> _logger;
+
+        public async Task<LdapSettings> GetSettings()
+        {
+            return await _ldapSettingsService.GetSettingsAsync();
+        }
+
+        public async Task<OmbiUser> LdapEntryToOmbiUser(LdapEntry entry)
+        {
+            var settings = await GetSettings();
+            var userName = GetLdapAttribute(entry, settings.UsernameAttribute).StringValue;
+
+            return new OmbiUser
+            {
+                UserType = UserType.LdapUser,
+                ProviderUserId = entry.Dn,
+                UserName = userName
+            };
+        }
+
+        private LdapAttribute GetLdapAttribute(LdapEntry userEntry, string attr)
+        {
+            try
+            {
+                return userEntry.GetAttribute(attr);
+            }
+            catch (Exception)
+            {
+                return null;
+            }
+        }
+
+        private async Task<LdapConnection> BindLdapConnection(string username, string password)
+        {
+            var settings = await GetSettings();
+
+            var ldapClient = new LdapConnection { SecureSocketLayer = settings.UseSsl };
+            try
+            {
+                if (settings.SkipSslVerify)
+                {
+                    ldapClient.UserDefinedServerCertValidationDelegate += LdapClient_UserDefinedServerCertValidationDelegate;
+                }
+
+                ldapClient.Connect(settings.Hostname, settings.Port);
+                if (settings.UseStartTls)
+                {
+                    ldapClient.StartTls();
+                }
+
+                ldapClient.Bind(username, password);
+            }
+            catch (Exception e)
+            {
+                _logger.LogError(e, "Failed to Connect or Bind to server");
+                throw e;
+            }
+            finally
+            {
+                ldapClient.UserDefinedServerCertValidationDelegate -= LdapClient_UserDefinedServerCertValidationDelegate;
+            }
+
+            if (!ldapClient.Bound)
+            {
+                ldapClient.Dispose();
+                return null;
+            }
+
+            return ldapClient;
+        }
+
+        private async Task<ILdapSearchResults> SearchLdapUsers(LdapConnection ldapClient)
+        {
+            var settings = await GetSettings();
+
+            string[] searchAttributes = { settings.UsernameAttribute };
+
+            _logger.LogDebug("Search: {1} {2} @ {3}", settings.BaseDn, settings.SearchFilter, settings.Hostname);
+            return ldapClient.Search(settings.BaseDn, LdapConnection.ScopeSub, settings.SearchFilter, searchAttributes, false);
+        }
+
+        public async Task<ILdapSearchResults> GetLdapUsers()
+        {
+            var settings = await GetSettings();
+
+            using var ldapClient = await BindLdapConnection(settings.BindUserDn, settings.BindUserPassword);
+
+            return await SearchLdapUsers(ldapClient);
+        }
+
+        public async Task<LdapEntry> LocateLdapUser(string username)
+        {
+            var settings = await GetSettings();
+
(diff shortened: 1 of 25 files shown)


src/Ombi.Core/Authentication/LdapUserManager.cs
```diff
@@ -0,0 +1,168 @@
...
+
+        public async Task<LdapSettings> GetSettings()
+        {
+            return await _ldapSettingsService.GetSettingsAsync();
+        }
+
+        public async Task<OmbiUser> LdapEntryToOmbiUser(LdapEntry entry)
+        {
+            var settings = await GetSettings();
+            var userName = GetLdapAttribute(entry, settings.UsernameAttribute).StringValue;
```
tidusjar, line 33: This is going to throw a `NullReferenceException` by calling `.StringValue` on a null object

src/Ombi.Core/Authentication/LdapUserManager.cs
```diff
@@ -0,0 +1,168 @@
...
+                ldapClient.Dispose();
+                return null;
+            }
+
+            return ldapClient;
+        }
+
+        private async Task<ILdapSearchResults> SearchLdapUsers(LdapConnection ldapClient)
+        {
+            var settings = await GetSettings();
```
tidusjar, line 96: We should also just pass the settings object into this method, saves another DB call

src/Ombi.Core/Authentication/LdapUserManager.cs
```diff
@@ -0,0 +1,168 @@
...
+
+            string[] searchAttributes = { settings.UsernameAttribute };
+
+            _logger.LogDebug("Search: {1} {2} @ {3}", settings.BaseDn, settings.SearchFilter, settings.Hostname);
+            return ldapClient.Search(settings.BaseDn, LdapConnection.ScopeSub, settings.SearchFilter, searchAttributes, false);
```
tidusjar, line 101: It's possible for `ldapClient` to be null here

src/Ombi.Core/Authentication/LdapUserManager.cs
```diff
@@ -0,0 +1,168 @@
...
+            if (ldapUsers == null)
+            {
+                return null;
+            }
+
+            while (ldapUsers.HasMore())
+            {
+                var currentUser = ldapUsers.Next();
+                var foundUsername = GetLdapAttribute(currentUser, settings.UsernameAttribute)?.StringValue;
+                if (foundUsername == username)
```
tidusjar, line 128: It would be better to do a `string.Equals` and use a case insensitive pattern match, see here: https://docs.microsoft.com/en-us/dotnet/api/system.string.equals?view=netcore-3.1#System_String_Equals_System_String_System_String_System_StringComparison_

src/Ombi.Core/Authentication/LdapUserManager.cs
```diff
@@ -0,0 +1,168 @@
...
+            {
+                return false;
+            }
+
+            try
+            {
+                using var ldapClient = await BindLdapConnection(ldapUser.Dn, password);
+                return (bool)ldapClient?.Bound;
+            } catch (Exception)
+            {
```
tidusjar, line 156: Again we are potentially losing the exception here

src/Ombi.Core/Authentication/OmbiUserManager.cs
```diff
@@ -49,19 +51,52 @@ public class OmbiUserManager : UserManager<OmbiUser>
...
         }
 
         private readonly IPlexApi _plexApi;
         private readonly IEmbyApiFactory _embyApi;
+        private readonly ILdapUserManager _ldapUserManager;
         private readonly ISettingsService<EmbySettings> _embySettings;
         private readonly ISettingsService<AuthenticationSettings> _authSettings;
+        private readonly ISettingsService<UserManagementSettings> _userManagementSettings;
+
+        public async Task<OmbiUser> FindUser(string userName)
```
tidusjar, line 73: I think this method should be named a bit different, It's purpose is not to find the user, It's to create a user from ldap

src/Ombi.Schedule/Jobs/Ldap/LdapUserImporter.cs
```diff
@@ -0,0 +1,87 @@
...
+            await _notification.Clients.Clients(NotificationHub.AdminConnectionIds)
+                .SendAsync(NotificationHub.NotificationEvent, "LDAP User Importer Started");
+            var allUsers = await _userManager.Users.Where(x => x.UserType == UserType.LdapUser).ToListAsync();
+
+            var allLdapUsers = await _ldapUserManager.GetLdapUsers();
+
+            while (allLdapUsers.HasMore())
+            {
+                var currentUser = allLdapUsers.Next();
+                var existingEmbyUser = allUsers.FirstOrDefault(x => x.ProviderUserId == currentUser.Dn);
```
tidusjar, line 55: please can we rename this variable from `existingEmbyUser`

src/Ombi.Settings/Settings/Models/LdapSettings.cs
```diff
@@ -0,0 +1,81 @@
+﻿namespace Ombi.Settings.Settings.Models
+{
+    public class LdapSettings : Settings
+    {
+        public LdapSettings()
+        {
+            IsEnabled = false;
```
tidusjar, line 7: Can we please remove this constructor, if we want to assist the user and pre-populate the settings then we should do that in the UI

src/Ombi.Core/Authentication/OmbiUserManager.cs
tidusjar: Oh I see what you have done here. In that case can we extract the ldap stuff into it's own private method? I can see the usecase where we will also now do this for Plex & emby users

tidusjar (changes requested):
Nice work so far! I have yet to get a LDAP server up and running, but I'll try and do that asap to test this.

Also the backend build is failing due to some unit tests now failing

dpraul:
Whewps guess I should return to this PR soon!

joshuaboniface:
I'm interested in testing this out too, so I'll give some feedback on how well it works with OpenLDAP and in comparison to Jellyfin-"proxied" imports over the next week or so. I did a bunch of work on the LDAP Jellyfin plugin so I can hopefully catch some things. One comment: the `UserType` of 5 is a conflict with my Jellyfin PR #3924; I don't mind changing one or the other, if you want to lay claim to it ;-)

dpraul:
Gonna try to actually return to this PR soon!

To that extent, I've since nuked my C# setup - which was no loss since I'm a baby C# developer and had no clue what I was doing. Is there any documentation I missed on setting up the development workspace for this repo? I had a messy time before swapping between vscode/WebStorm for the TypeScript portions and Visual Studio for the C# portions.

> I'm interested in testing this out too, so I'll give some feedback on how well it works with OpenLDAP and in comparison to Jellyfin-"proxied" imports over the next week or so. I did a bunch of work on the LDAP Jellyfin plugin so I can hopefully catch some things. One comment: the `UserType` of 5 is a conflict with my Jellyfin PR #3924; I don't mind changing one or the other, if you want to lay claim to it ;-)

Did you ever do any testing? I've had this running pretty much since I created this PR with no issues using OpenLDAP, but I have very few users, so would love more UATing. Also - needless to say at this point, but you got `UserType` 5 in before I did, so I will update this.

BurninTurtles:
Hi All! If there is still any interest in this PR, I would love to be a tester!

PR #3817, opened 2020-10-17
what do you think?

goldenpipes:
basically got rid of some extra padding to the layout change buttons. and hid them on mobile since they it doesn't really need to be there on mobile.

image

let me know.

### DIFF
--- a/src/Ombi/ClientApp/src/app/discover/components/discover/discover.component.html
+++ b/src/Ombi/ClientApp/src/app/discover/components/discover/discover.component.html
@@ -18,7 +18,7 @@
     </div>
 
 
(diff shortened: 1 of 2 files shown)


twanariens:
Well you definately got rid of some padding :D But. Would love to see a little bit of padding between the Movies/Combined/TV bar and the Popular/Trending/Upcoming bar.

And how does it hold up when you resize your screen? (like 2560 x 1440)

goldenpipes:
1440p SCREENSHOT

good catch on the padding between the selector's forgot to add that back in

image

twanariens:
Look amazing. Now if you can try and center the search bar. We should make some users happy :D

PR #3840, opened 2020-10-29
Update PlexContentSync.cs

Berserkir-Wolf:
Fix incorrect logging and notification message


tidusjar:
The content variable is a combination of TV shows and Movies, it's not just movies :(

PR #3904, opened 2020-12-03
Fixed: Don't Type RadarrAddMovieResponse Image list to String

Qstick:
In both the v2 and v3 Radarr API responses, the images property of a movie is a list of Image objects. This has never been a list of strings and causes errors with Ombi V3 and Radarr V3 given V3 returns all metadata on add, V2 always returned an empty list on a movie add (of the same image object as V3). 

It seems this has been an issue since the beginning in Ombi, but wasn't noticed with V2. 

Option 2 just remove it given its not used anyway, we are getting user complaints for those that don't want to use v4 due to bugs... can we please backport this and release, its a couple lines.


tidusjar:
Thanks!

PR #3924, opened 2020-12-10
Add separate Jellyfin server type

joshuaboniface:
Due to forthcoming changes to the Jellyfin API, this PR adds support for
Jellyfin as server type completely independent from Emby. It also undoes
the workarounds that treated Jellyfin as a subset of Emby.

This does add a decent amount of code duplication today, however it sets
the stage for Jellyfin to diverge from Emby arbitrarily without concern for
hacks or workarounds to try to support this. Especially as we continue
with our database and API rewrites, this is going to become more and
more important.

This PR is still in draft state. The first commit handles my attempt to add
Jellyfin everywhere. Basically anywhere I found an "Emby" reference, I
added a Jellyfin one below/after it. It does build properly as far as I can
tell, and is installable and runnable.

Still to do:

- [x] Add database migration ~(currently unable to run, help has been requested on Discord, can't test until I do)~ ~(have run, but doesn't work for some reason)~.
- [x] Test this thoroughly with my own instance.


src/Ombi.Api.Jellyfin/JellyfinApiFactory.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Threading.Tasks;
+
+namespace Ombi.Api.Jellyfin
+{
+    public class JellyfinApiFactory : IJellyfinApiFactory
+    {
+        private readonly ISettingsService<JellyfinSettings> _jellyfinSettings;
+        private readonly IApi _api;
+
+        // TODO, if we need to derive futher, need to rework
```
Artiume, line 13:
```suggestion
        // TODO, if we need to derive further, need to rework
```

src/Ombi.Api.Jellyfin/Models/Media/JellyfinImagetags.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Ombi.Api.Jellyfin.Models.Movie
+{
+    public class JellyfinImagetags
+    {
+        public string Primary { get; set; }
+        public string Logo { get; set; }
+        public string Thumb { get; set; }
+
```
Artiume, line 8: space?

src/Ombi.Api.Jellyfin/JellyfinApiFactory.cs
joshuaboniface: A lot of this was straight copy+paste+replace, so a lot of these are not mine.

src/Ombi.Api.Jellyfin/Models/Media/JellyfinImagetags.cs
joshuaboniface: ibid.

src/Ombi.Api.Jellyfin/JellyfinApi.cs
```diff
@@ -67,15 +67,15 @@ public async Task<EmbyUser> LogIn(string username, string password, string apiKe
 
             request.AddJsonBody(body);
 
-            request.AddHeader("X-Emby-Authorization",
+            request.AddHeader("X-Jellyfin-Authorization",
```
crobibero, line 70:
```suggestion
            request.AddHeader("X-Emby-Authorization",
```

We don't currently have `X-Jellyfin-Authorization`

joshuaboniface: Aah so it is still that, I wasn't 100% so just went with it.

joshuaboniface:
Some initial testing results.

Despite the migration failure to apply (I just forced it through), this are generally working. It's able to connect to my Jellyfin 10.7.0-rc1 instance and seems to be gathering stats. I'll try adding some more things to make sure it actually works but no obvious errors were thrown.

The user importer needs some work, I thought I added that but it turns out it's not showing up so I missed something.

I also need to figure out why MySQL migrations aren't being run (I suspect I need to connect it to a real DB) though it should be identical to the SQLite one as well.

joshuaboniface:
I think I've got all the code components, fixed several issues and missed fields, as well as added some additional information especially to `migration.txt` (now called `migration.md` so it will format in GitHub).

Migrations for both SQLite and MySQL have been added and, while right now I can only confirm the SQLite migration, both *should* work.

The user importer now works perfectly as well.

This *should* be very close to completion, now proceeding with further testing.

joshuaboniface:
Ready for review. All tests except the security test are passing (but I can't see why that one doesn't). So far all my testing has been successful and everything looks to be working as expected.

tidusjar:
Wow, nice work! Will review in the next few days! Thanks for your work on this.

src/Ombi.Api.Jellyfin/JellyfinApiFactory.cs
tidusjar: That's fine. I'll sort it

tidusjar:
Going to pull into a separate branch and fix up the conflicts and do some testing, thanks a lot for all this work again!

PR #3932, opened 2020-12-13
jobs page and the update page

goldenpipes:


goldenpipes:
been awhile 😃 

I know the updater is still broken, but at least the page is styled.

PR #3970, opened 2021-01-10
Create zh.json

freefrank:
Chinese translation.


tidusjar:
Thanks, I need to add some more into ombi other than just the translation file, but this is a start!

freefrank:
Thank you for this greate project! I'll continue tweak the translation. 
Also, there are two language code for Chinese, zh_CN and zh_TW, would ombi able to pick up file with more than 2 characters in filename?

PR #4022, opened 2021-02-02
Provide the base URL if possible for Sonarr/Radarr

jackodsteel:
If the API returns a base URL that differs from the one currently set, output that in the error message.

Currently we show that the test passed, which is definitely wrong, as even though Sonarr/Radarr return a result here, they will fail for POST requests if the base URL is not set correctly, which is pretty confusing.

### DIFF
--- a/src/Ombi/ClientApp/src/app/settings/lidarr/lidarr.component.ts
+++ b/src/Ombi/ClientApp/src/app/settings/lidarr/lidarr.component.ts
@@ -1,4 +1,4 @@
-﻿import { Component, OnInit } from "@angular/core";
+import { Component, OnInit } from "@angular/core";
 import { FormBuilder, FormGroup, Validators } from "@angular/forms";
 
 import { ILidarrSettings, IMinimumAvailability, IProfiles, IRadarrProfile, IRadarrRootFolder } from "../../interfaces";
@@ -106,11 +106,13 @@ export class LidarrComponent implements OnInit {
             return;
         }
         const settings = <ILidarrSettings>form.value;
-        this.testerService.lidarrTest(settings).subscribe(x => {
-            if (x === true) {
-                this.notificationService.success("Successfully connected to Lidarr!");
+        this.testerService.lidarrTest(settings).subscribe(result => {
+            if (result.isValid) {
+                this.notificationService.success("Successfully connected to Sonarr!");
+            } else if (result.expectedSubDir !== null) {
+                this.notificationService.error("Your Sonarr Base URL must be set to " + result.expectedSubDir);
             } else {
-                this.notificationService.error("We could not connect to Lidarr!");
+                this.notificationService.error("We could not connect to Sonarr!");
             }
         });
     }
--- a/src/Ombi/ClientApp/src/app/settings/lidarr/lidarr.component.ts
+++ b/src/Ombi/ClientApp/src/app/settings/lidarr/lidarr.component.ts
@@ -108,11 +108,11 @@ export class LidarrComponent implements OnInit {
         const settings = <ILidarrSettings>form.value;
         this.testerService.lidarrTest(settings).subscribe(result => {
             if (result.isValid) {
-                this.notificationService.success("Successfully connected to Sonarr!");
+                this.notificationService.success("Successfully connected to Lidarr!");
             } else if (result.expectedSubDir !== null) {
-                this.notificationService.error("Your Sonarr Base URL must be set to " + result.expectedSubDir);
+                this.notificationService.error("Your Lidarr Base URL must be set to " + result.expectedSubDir);
             } else {
-                this.notificationService.error("We could not connect to Sonarr!");
+                this.notificationService.error("We could not connect to Lidarr!");
             }
         });
     }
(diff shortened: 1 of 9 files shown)


tidusjar:
Thanks for the contribution!

Correct me if I am wrong, if they are trying to test the connection and the base URL is incorrect, we will not be able to get the system information because that's the wrong endpoint?

src/Ombi/ClientApp/src/app/settings/lidarr/lidarr.component.ts
```diff
@@ -106,11 +106,13 @@ export class LidarrComponent implements OnInit {
...
-            if (x === true) {
-                this.notificationService.success("Successfully connected to Lidarr!");
+        this.testerService.lidarrTest(settings).subscribe(result => {
+            if (result.isValid) {
+                this.notificationService.success("Successfully connected to Sonarr!");
+            } else if (result.expectedSubDir !== null) {
+                this.notificationService.error("Your Sonarr Base URL must be set to " + result.expectedSubDir);
             } else {
-                this.notificationService.error("We could not connect to Lidarr!");
+                this.notificationService.error("We could not connect to Sonarr!");
```
tidusjar, line 115: Please change back

src/Ombi/ClientApp/src/app/settings/radarr/radarr.component.ts
```diff
@@ -96,11 +96,13 @@ export class RadarrComponent implements OnInit {
...
-            if (x === true) {
-                this.notificationService.success("Successfully connected to Radarr!");
+        this.testerService.radarrTest(settings).subscribe(result => {
+            if (result.isValid) {
+                this.notificationService.success("Successfully connected to Sonarr!");
+            } else if (result.expectedSubDir !== null) {
+                this.notificationService.error("Your Sonarr Base URL must be set to " + result.expectedSubDir);
             } else {
-                this.notificationService.error("We could not connect to Radarr!");
+                this.notificationService.error("We could not connect to Sonarr!");
```
tidusjar, line 105: Please change back

src/Ombi/ClientApp/src/app/settings/radarr/radarr.component.ts
```diff
@@ -96,11 +96,13 @@ export class RadarrComponent implements OnInit {
             return;
         }
         const settings = <IRadarrSettings> form.value;
-        this.testerService.radarrTest(settings).subscribe(x => {
-            if (x === true) {
-                this.notificationService.success("Successfully connected to Radarr!");
+        this.testerService.radarrTest(settings).subscribe(result => {
+            if (result.isValid) {
+                this.notificationService.success("Successfully connected to Sonarr!");
```
tidusjar, line 101: Please change back

src/Ombi/ClientApp/src/app/settings/lidarr/lidarr.component.ts
```diff
@@ -106,11 +106,13 @@ export class LidarrComponent implements OnInit {
             return;
         }
         const settings = <ILidarrSettings>form.value;
-        this.testerService.lidarrTest(settings).subscribe(x => {
-            if (x === true) {
-                this.notificationService.success("Successfully connected to Lidarr!");
+        this.testerService.lidarrTest(settings).subscribe(result => {
+            if (result.isValid) {
+                this.notificationService.success("Successfully connected to Sonarr!");
```
tidusjar, line 111: Please change back

src/Ombi/ClientApp/src/app/settings/lidarr/lidarr.component.ts
```diff
@@ -106,11 +106,13 @@ export class LidarrComponent implements OnInit {
...
+            } else if (result.expectedSubDir !== null) {
+                this.notificationService.error("Your Sonarr Base URL must be set to " + result.expectedSubDir);
```
tidusjar, line 113: "Lidarr"

src/Ombi/ClientApp/src/app/settings/radarr/radarr.component.ts
```diff
@@ -96,11 +96,13 @@ export class RadarrComponent implements OnInit {
...
+            } else if (result.expectedSubDir !== null) {
+                this.notificationService.error("Your Sonarr Base URL must be set to " + result.expectedSubDir);
```
tidusjar, line 103: "Radarr"

tidusjar (changes requested):

src/Ombi/ClientApp/src/app/settings/lidarr/lidarr.component.ts
jackodsteel: 🤦‍♂️ I must not copy paste at 4am! Fixed

jackodsteel:
Sorry about the copy paste errors!

> Correct me if I am wrong, if they are trying to test the connection and the base URL is incorrect, we will not be able to get the system information because that's the wrong endpoint?

Well, yes and no. If it's the _wrong_ base URL, it will fail. But if there's no base URL specified, they send a 302 redirect to the link with the base URL, which the .NET HttpClient follows, which then returns the expected result. So if you should have a base URL, but don't specify one, it shows as successful (I ran into this).
The annoying part, is that although this works for GET requests, it doesn't for the POSTs that Ombi tries to issue once you actually request content, so those fail, even though the test appeared to work.

My first take at fixing this just set the HttpClient to not follow redirects, which did cause the test to fail, so was a suitable fix, but this way ends up being more informative so went this way.

---

Also worth noting that this problem probably exists for any other app that sends a redirect for the GET requests we issue in the Tester, but then fails in the POSTs. Haven't tried other apps though, but worth keeping in mind for future bug reports or extra testing.

tidusjar:
Thanks for the response! Just tested this and it's working nicely! Nice one for this, much appreciated!

tidusjar (approved):

Issue #3194, opened 2019-10-09
Sonarr - Ombi send wrong ID for TV Shows -Bad Request: 0 is not a valid ID


GhostPyxel:
**Describe the bug**
When trying to add a new TV Show OR missing episodes of existing series, Sonarr report an invalid ID from Ombi and logs in failed requests. The main error reporting here is an Invalid Root Path.

With a bit of searching, I think Sonarr is "partially" detected in Ombi. (More explanation on that through the report). Finally, this issue can be reproduce either on an admin account or a standard user account with no "Auto-Approve" role.

**To Reproduce**
Sonarr settings in Ombi:

- Enable 
- No V3
- Sonarr Local IP
- Port
- Sonarr API Key
- No SSL and no BaseURL

Load Qualities (SUCCESS and CORRECT)
Load Folders (SUCCESS and CORRECT)
Test Connectivity (SUCCESS)

1. Search a TV Show in search bar.
2. Request any Season or Episode.
3. Sonarr report 3 error logs and can't proceed. 

**Expected behavior**
Ombi should send a correct TV Show ID to Sonarr.

**Screenshots**
Here's why I say Sonarr is "partially" detected by Ombi. I've notice that I'm unable to go on PLEX through Ombi on the TV Show section. On the movies, I have this layout:

![movies](https://user-images.githubusercontent.com/52525576/66524925-1d89ca00-eaf4-11e9-8310-0d016ed16bff.jpg)

As you can see, summary of the movies are translated in my language and the button "Regarder sur Plex" (See on Plex) is here ect.. Link of each movie works and redirect on Plex correct movie. However in the TV Show tab, I have this:

![TV Shows](https://user-images.githubusercontent.com/52525576/66525035-6f325480-eaf4-11e9-8c51-9bc804962ed7.jpg)

No translation, no similar series, no link to Plex. And many TV shows are missing and they are not registred as "Available". I can try to requests the missing shows but Ombi respond "TV Show already existing on drive, cannot add it to the requests list. Ombi detects the TV Show existing but don't update it through the UI.

**Logs (Logs directory where Ombi is located)**
Major list  of request failing.
```
2019-10-10 23:29:52.434 [Error] StatusCode: BadRequest, Reason: Bad Request, RequestUri: http://localhost:8989/sonarr/api/series/
2019-10-10 23:29:55.178 [Error] StatusCode: BadRequest, Reason: Bad Request, RequestUri: http://localhost:8989/sonarr/api/series/
[...]

PR #4116, opened 2021-03-28
Verify Sonarr quality/rootfolder override

kitzin:
Ombi implicitly trusts that the quality/rootfolder override is set correctly, and causes issues downstream.

Line `258` in [TvSender.cs](https://github.com/Ombi-app/Ombi/tree/develop/src/Ombi.Core/Senders/TvSender.cs#L258) it tries to add a series, but the `AddSeries` function in [SonarrApi.cs](https://github.com/Ombi-app/Ombi/blob/develop/src/Ombi.Api.Sonarr/SonarrApi.cs#L92) fails to validate the `NewSeries` object since the `qualityProfileId` attribute is set to `0`. In the end it assumes the series is added in Sonarr and causes a API request with series id set to `0` which causes the following error.

```
[Error] StatusCode: BadRequest, Reason: Bad Request, RequestUri: http://SONARR_HOST/sonarr/api/series/0
```

#3194 could be related to this.

### DIFF
--- a/src/Ombi.Core/Senders/TvSender.cs
+++ b/src/Ombi.Core/Senders/TvSender.cs
@@ -206,7 +206,11 @@ namespace Ombi.Core.Senders
             // Overrides on the request take priority
             if (model.ParentRequest.QualityOverride.HasValue)
             {
-                qualityToUse = model.ParentRequest.QualityOverride.Value;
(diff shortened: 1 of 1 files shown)


tidusjar:
Nice work! thanks for this!

I think you are correct that it's related to that bug, but the new advanced options had made the bug more apparent. So thanks!

PR #4122, opened 2021-03-28
fixes rootfolder selection in sonarr

kitzin:
I realized that in #4116 I overlooked that it always sets the `RootFolder` in line `171` and `191` so the rootfolder check further down does nothing.


tidusjar:
Nice find. 

Thanks!

PR #4132, opened 2021-03-31
Develop kuraki

twanariens:
Some styling changes and a rework of the top-container (search)


twanariens:
DO NOT accept this PR. It has a bug

PR #4150, opened 2021-04-17
Fix TMDB logo URL

mattmattmatt:
Previous value `{{baseUrl}}images/tmdb-logo.svg` leads to broken image requests, e.g. `server.com/ombiimage/tmdb-logo.svg`. We need to trail the `baseURL` with a `/` to ensure the correct path, e.g. `server.com/ombi/image/tmdb-logo.svg`. See also lines 9 and lines 14.


tidusjar:
Thanks!

PR #4193, opened 2021-05-18
Autofocus the Search field

AbeKline:
So on page load a user can immediately start typing a search without having to manually navigate to it. Also a big bonus for accessibility


tidusjar:
Thanks!

Nice idea

AbeKline:
Did that actually work? If so we should probably resolve this 
https://ombifeatures.featureupvote.com/suggestions/186038/autofocus-search-bar-on-load-of-landing-page

PR #4199, opened 2021-05-19
Filter out Genres from Discover

Ashton-Sidhu:
Hi @tidusjar - opening up this PR now for discussion + feedback.

This is the PR to introduce filtering out genres from the Discover lists.

TheMovieDB API only has the capability to get a list of all genre mappings and unfortunately does not have a lookup api

### DIFF
--- a/src/Ombi.Settings/Settings/Models/External/TheMovieDbSettings.cs
+++ b/src/Ombi.Settings/Settings/Models/External/TheMovieDbSettings.cs
@@ -7,5 +7,9 @@ namespace Ombi.Core.Settings.Models.External
         public bool ShowAdultMovies { get; set; }
 
         public List<int> ExcludedKeywordIds { get; set; }
+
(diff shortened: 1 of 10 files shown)


Ashton-Sidhu:
I've added the back end API calls to get the genres and communicate them with the front end.

For the UI, since there are only ~15 or so genres I'm thinking of making the Genre exclusion a drop down instead of a text box, let me know your thoughts!

tidusjar:
Wow that was quick! I'll get onto reviewing asap

tidusjar:
Also I'm fine with making it a drop-down, better for accessibility with such a small set of data

Ashton-Sidhu:
Ok thanks, I'll turn it into a drop down - the component init code will change slightly then as well. Will work on it and push the changes soon.

tidusjar:
Code looks good! I'll do some testing!

tidusjar:
Nice work! All functioning correctly! Thanks a lot for this!

PR #4201, opened 2021-05-20
Update user-preference.component.html

jamesmacwhite:
Fixes mixture of quotes on iOS and Android app image links and removes px from width attribute as this is not valid.


tidusjar:
Nice! Thanks for this

PR #4221, opened 2021-06-08
Smaller Side-nav for better experience on discover page fix #4219

### LINKED ISSUE #4219: V4 Sidebar too wide
Sidebar is way too wide compared to popular interfaces such as Radarr. Sonarr, Plex etc.  Currently, OMBI's sidebar wastes too much space.**  This wasted space could be used for posters instead.  

Please consider using single-word categories.  I.E., `User Management` should be just `Users`.  `Feature suggestions` renamed to `Ideas` or something like that (similar to other apps)
Wronte

**Below, looks nicer; and, wastes less space.  This is just a rough idea.. something like this:**
Correct

twanariens:


tidusjar:
One problem I see is when scrolling the carousel when in a smaller view (e.g. mobile) then it can cut off the items:
image
image

Other than that, it looks good!

I've also fixed the UI Automation tests in this PR

tidusjar (changes requested):
See discussion

mkanet:
@tidusjar sorry to intrude.  You just reminded me... I just wanted to point out that currently, Ombi's poster sizes are WAY too large in respect to similar web interfaces (Sonarr, Radarr, etc).  It doesn't look like this PR addresses the poster sizes being too large.  Maybe you guys would please look into why the posters are so large?

**Ombi:**
Ombi

**Sonarr:**
Sonarr

**Radarr:**
Radarr

twanariens:
Cheated a bit. The cutoff keeps happening since I stated a fixed width for each poster (to keep dimensions for the picture).
So the carousel keeps on populating until the full screen width is filled. This includes posters which does not actually fit (to fill the page width).

I have changed that for low width devices (no laptops or pc's) the fixed width is not set but I removed the fixed height. Meaning the ratio will remain but the poster can be bigger then they were.

For pc and laptops, I changed the amount of posters scrolled when you click the arrows. Meaning the cut off poster will show fully when you navigate with the arrows.

tidusjar:
Looking good!

tidusjar (approved):

PR #4273, opened 2021-07-28
Included row bulk selection from Movie Request List into TV request list

ThePheonixGuy:
This includes the row selectors from the Movie Request list into the TV show request list. 

However, I have been unable to get the app to run locally - when running the Ombi project under Kestrel, it does not build and copy the client app and the SpaProxy is unable to proxy requests. I am not sure if I am running things incorrectly?

If someone could guide me on that, I'll happily do so and update this PR.

### DIFF
--- a/src/Ombi/ClientApp/src/app/requests-list/components/tv-grid/tv-grid.component.html
+++ b/src/Ombi/ClientApp/src/app/requests-list/components/tv-grid/tv-grid.component.html
@@ -30,6 +30,20 @@
 
     <table mat-table [dataSource]="dataSource" class="requests table" matSort [matSortActive]="defaultSort" matSortDisableClear [matSortDirection]="defaultOrder">
 
+      <ng-container matColumnDef="select" *ngIf="isAdmin">
+        <th  mat-header-cell *matHeaderCellDef>
+          <mat-checkbox id="adminMasterCheckbox" (change)="$event ? masterToggle() : null"
+                        [checked]="selection.hasValue() && isAllSelected()"
+                        [indeterminate]="selection.hasValue() && !isAllSelected()">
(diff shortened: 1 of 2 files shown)


tidusjar:
Sorry! I've only just seen this. To get the app to run locally you can follow this: https://docs.ombi.app/development/building-ombi/

tidusjar:
So, it seems that we need to add the following into the NgOnInit to get the checkboxes to show as it's not defined in the `displayedColumns`

```
        if (this.isAdmin) {
            this.displayedColumns.unshift('select');
        }
```

The main reason why I didn't do this to start is that a TV show can have many individual requests, so by doing the bulk, it needs to go through each Tv Request for that show and either approve or delete it. Keeping in mind that some might be already available e.g.

Request for the TV Show Dexter

User A requests season 1
Season 2 is available from a previous request
User B Requests Season 3 and has been approved


The above has 3 different requests for the same TV show in completely different states. So i fear that introducing this functionality might confuse users and lead to potential mistakes.

ThePheonixGuy:
I see your point WRT confusion. The user experience would make it a bit more complicated. The potential outcome here would have the rows be expandable and then having multi selects in their for bulk approval. 

Let me know if you want me to continue work on this or rather close it due to the user experience? Happy either way - will maintain a fork of the frontend for my purposes if we'd prefer not to merge this in 👍🏼

tidusjar:
I'm happy for it to expand and then list them there with the multi select. What you can probably do is just use the same component that's on the details page and just modify it :)

ThePheonixGuy:
Ooh! I forgot about that component on the details page. Yeah I will look into doing that. Going to close this and reopen when I have that available.

PR #4304, opened 2021-08-29
Updating bulgarian translation file

Vbgf:
I've updated the bulgarian translation file based on the en.json translation and previous work on the bg.json file by my peers.

Also fixed a minor typo in the en.json file where "AirDate" was replaced by "Air Date".


tidusjar:
Thanks!

PR #4320, opened 2021-10-04
Updates short date formatting to use user's browser preference.

bernarden:
Formatting for all short dates is now using window.navigator.language. 
Dates for chat messages are also converted from UTC to local time first.
image

Let me know if you think using window.navigator.languages[0] (if present) might be a better option in your opinion.


tidusjar:
Looks good, I like your approach. Nice work!

Issue #4263, opened 2021-07-23
'Request' button still shown in Search results and Collections view for already Requested items


ghost:
**Describe the bug**
When viewing a movie Collection, an Actors (list of movies they're in), or the results of a search, items that have already been requested still give the option to Request it under a normal user (and admin).

**To Reproduce**
For Searching:
1. Search for F9 (or some other movie not already Available)
2. Click on "Request"
3. Search again F9
4. The 'Request' button still exists
5. Confirm F9 is already requested by clicking into F9 details

and

For Movie Collections:
1. For the Fast & Furious movie collection, find and click on F9 movie
2. Click on 'Request' button (if not already requested)
3. Click on 'Collection View'
4. Hover over movie 'F9'
5. See Issue - Movie can still be requested, however gives error to say its already been requested.

or 

1. For the Fast & Furious movie collection, find and click on F9 movie
2. Click on 'Collection View'
4. Click on 'Request Collection'
5. Wait for collection items to be marked as 'Requested'
5. Hover over movie 'F9'
5. See Issue - Movie can still be requested, however gives error to say its already been requested.

etc...

**Expected behavior**
The search results, actor list of movies/shows and collection views should be use to easily identify which items have already been requested. 

1. Once a movie has been requested by the *current* (normal) user, the option to request it again should not be available in search results (or Collections view). At the moment, the 'Request' button still appears in Search results and Collection View. This button should be removed, and a 'Requested' status should appear instead (aka Pending?).
- ie Remove the 'Request' button and have a 'Requested/Pending' status for already requested items *by the user* (eg similarly displayed as the 'Available' status, only with a yellow dot. The 'Available' status should also be changed to a dark green dot to assist people who are colour impaired to tell the difference for ESL, etc.), and

2. For movies already requested by *other users*, the 'Request' button in search results (or Collections view) could change to 'Requested/Pending'. This will allow the current user to still add request weighting to the existing request (or ignore re-requesting it),  since they can easily see its already been requested.
- ie If config option 'Hide Requests from Other Users' is **_disabled_**, then also change the 'Request' button to a 'Requested/Pending' button for items already requested *by other users*, otherwise if the option to 'Hide Other User requests' is **_enabled_**, then leave this as is (eg 'Request' button shown).

**Please note**: The current function (way it works) should not be changed for Admin and Power users - just change the button label to 'Requested/Pending' and still have the same function for admin/pu's (ie still allow admin/pu's to add another request of the same item,etc.).

[...]

PR #4323, opened 2021-10-05
Removes overwrite of requested field with undefined.

bernarden:
Resolves the #4263 issue.


tidusjar:
Nice spot! Thanks again :)

PR #4359, opened 2021-10-19
Localization

sephrat:
This should cover ~95% of the screens to the end user (not admin). 
Known unlocalized stuff:
- error (and possibly success) messages which are hardcoded on the backend ==> I may tackle this later but consider it low prio. 
- Series and Movies titles in Requests view are not localized, whereas they are in the Discover panels. I don't know my way around the code enough yet to have it saved in the DB in the correct language or have it fetched on the fly from the proper APIs. Help needed :-)

One thing worth of attention: since the release status is (I assume) fetched from 3rd party providers (TVDB/IMDB?), I translate it on the fly on the frontend. It's initialized with the statuses I could find manually, there may be some missing (see 'StatusValues' in en.json) but it will fallback to the current behavior if the status is not in the list so we should be fine.


tidusjar:
Thanks! 

I'll take a look into this soon :)

sephrat:
It's quite up to date, but I've just merged again, let's see. 
I'll try to figure out how to use cypress on my workstation to better understand what's wrong.

tidusjar:
Nice work!

PR #4366, opened 2021-10-21
Localization - step 2

sephrat:
- [x] Localize TV and movies titles in requests names
- [x] Localize newsletter
  - [x] Localize TV shows
  - [x] ~~Localize hard-coded texts in HTML ("New movies", etc. ) 🆘 help needed - is there any existing tool on the backend I can reuse to handle translations? If it's not the case, I'm considering the use of settings for such texts, like it was done for the subject and the leading text~~ Postponed to a later PR until I find a clean way to do it.
- [x] Localize music (as far as Ombi is concerned - Lidarr does not seem to support languages which will be the main issue)
- [x] Add a few missing release statuses
- [x] Localize error messages (use error codes on backend side and interpret them on frontend to localize them)
- [x] Minor fixes


src/Ombi.Core/Engine/RequestEngineResult.cs
```diff
@@ -1,11 +1,31 @@
...
         public string Message { get; set; }
         public bool IsError => !string.IsNullOrEmpty(ErrorMessage);
         public string ErrorMessage { get; set; }
+        public ErrorCode? ErrorCode { get; set; }
         public int RequestId { get; set; }
     }
+
+    [JsonConverter(typeof(StringEnumConverter))]
+    public enum ErrorCode {
+        AlreadyRequested,
```
tidusjar, line 18: Can we please assign the explicit values to these, so we don't break things on reordering for example

sephrat: StringEnumConverter will return the name of the member, not an ID. Reordering won't break this.
image

src/Ombi/ClientApp/src/app/media-details/components/movie/movie-details.component.ts
```diff
@@ -92,7 +92,7 @@ export class MovieDetailsComponent {
             dialog.afterClosed().subscribe(async (result) => {
                 if (result) {
                     const requestResult = await this.requestService.requestMovie({ theMovieDbId: this.theMovidDbId,
-                        languageCode: null,
+                        languageCode: navigator.language,
```
tidusjar, line 95: Same here regarding using the `TranslateService`

src/Ombi/ClientApp/src/app/media-details/components/tv/panels/tv-request-grid/tv-request-grid.component.ts
```diff
@@ -43,7 +44,7 @@ export class TvRequestGridComponent {
 
         const viewModel = <ITvRequestViewModelV2>{
             firstSeason: this.tv.firstSeason, latestSeason: this.tv.latestSeason, requestAll: this.tv.requestAll, theMovieDbId: this.tv.id,
-            requestOnBehalf: null
+            requestOnBehalf: null, languageCode: navigator.language
```
tidusjar, line 47: Same here with using the `TranslateService`

src/Ombi/ClientApp/src/app/services/message.service.ts
```diff
@@ -18,4 +20,14 @@ export class MessageService {
             this.snackBar.open(message, "OK", this.config)
         }
     }
+    public sendRequestEngineResultError(result: IRequestEngineResult, action: string = "Ok") {
+        console.log(result.errorCode);
```
tidusjar, line 24: Can we please remove this

src/Ombi/ClientApp/src/app/shared/episode-request/episode-request.component.ts
```diff
@@ -45,7 +45,7 @@ export class EpisodeRequestComponent {
 
         const viewModel = <ITvRequestViewModelV2>{
             firstSeason: this.data.series.firstSeason, latestSeason: this.data.series.latestSeason, requestAll: this.data.series.requestAll, theMovieDbId: this.data.series.id,
-            requestOnBehalf: this.data.requestOnBehalf
+            requestOnBehalf: this.data.requestOnBehalf, languageCode: navigator.language
```
tidusjar, line 48: Please use the `TranslateService`

tidusjar (changes requested):

sephrat:
Thank you for your code review. All should be fixed, except the enum values: see my reply on the thread.

src/Ombi.Core/Engine/RequestEngineResult.cs
tidusjar: Ok! Didn't spot that :)

tidusjar:
Nice work! looks good!

tidusjar (approved):

PR #4388, opened 2021-11-09
Add TV shows to Actor discover page

sephrat:
image
For now the TV shows and/or Movies are displayed or not according to the filter set. Not sure if it's the best idea yet, but maybe the solution is to have a more organized page for the actor? I'm open to criticism.


tidusjar:
Sorry, what is it that we have done here?

sephrat:
Sorry, I should have better explained my change.
The actor page used to show only Movies the actor played in. This is confusing especially when you click on an actor from a TV show and you can't see the show you come from. With this PR, it also lists TV shows they were in.
You can filter if you want to display only Movies or only TV shows by using the search toggles. This part I'm not so sure about the UX, maybe we shouldn't filter at all?

tidusjar:
Ah ok! That explains it :) 

This is something I've been meaning to get around to, but as always there's always something else.

Regarding the filter, I think we should ignore it, I always treat the filter to affect the search results and only the search results, as you mentioned, I think it would be quite confusing otherwise.

sephrat:
> This is something I've been meaning to get around to, but as always there's always something else.

No problem, that's why PRs are a thing. :-)

> Regarding the filter, I think we should ignore it

Done!

src/Ombi/ClientApp/src/app/discover/components/actor/discover-actor.component.ts
```diff
@@ -24,24 +25,40 @@ export class DiscoverActorComponent {
...
+        });
+        searches.push(moviesSearch);
+
+        var TvSearch = this.searchService.getTvByActor(this.actorId);
+        TvSearch.subscribe(res => {
+            this.pushDiscoverResults(res.cast, RequestType.tvShow);
+        });
+        searches.push(TvSearch);
+
+        forkJoin(searches).subscribe(() => {
```
tidusjar, line 49: Might be better to do something like this?

```
combineLatest([
  this.searchService.getMoviesByActor(this.actorId),
  this.searchService.getTvByActor(this.actorId),
]).subscribe(
  ([movie, tv]) => {
   this.pushDiscoverResults(movie.cast, RequestType.movie);
   this.pushDiscoverResults(tv.cast, RequestType.tvShow);
  }
);
```

src/Ombi/Controllers/V2/SearchController.cs
```diff
@@ -395,6 +395,19 @@ public async Task<ActorCredits> GetMoviesByActor(int actorId)
...
+        /// Returns all the tv shows that is by the actor id 
+        /// </summary>
+        /// <param name="actorId">TheMovieDb Actor ID</param>
+        /// <returns></returns>
+        [HttpGet("actor/{actorId}/tv")]
+        [ProducesResponseType(StatusCodes.Status200OK)]
+        [ProducesDefaultResponseType]
+        public async Task<ActorCredits> GetTvByActor(int actorId)
+        {
+            return await _tvEngineV2.GetTvByActor(actorId, null);
```
tidusjar, line 409: Let's pass the API the current language?

Maybe just pass it as a query parameter? 

e.g. `actor/1/tv?lang=en`.
You can do something like

```
public async Task<ActorCredits> GetTvByActor([FromRoute] int actorId, [FromQuery] string lang)
```

tidusjar (changes requested):
Looks good!

src/Ombi/Controllers/V2/SearchController.cs
sephrat: Good catch, I never noticed. I copy-pasted this from getMoviesByActor, which means the current page in stable releases displays movies in English only.. It's something I overlooked in my localization PR. 
Maybe it's best if I take this on in another PR, along with the movies part?

tidusjar: Yeah I'm fine for you to do that

sephrat: @tidusjar Mind if I ask why not use DefaultLanguageCode() instead of a query parameter? It seems to be used everywhere else in Ombi for displaying discover cards and I don't see the use case for it. See https://github.com/sephrat/Ombi/blob/0c4bd10b0f002fd67b55a07247828a7f40682433/src/Ombi.Core/Engine/MovieSearchEngine.cs#L126 for example.

tidusjar: Good point, `DefaultLanguageCode` reads from the user so it's fine then :)

PR #4397, opened 2021-11-10
Secure role "Manage Own Requests"

### LINKED ISSUE #4391: "Manage Own Requests" role allows to delete another user's request
**Describe the bug**
Any user with "Manage Own Requests" can delete anyone's requests.
The option is not visible in the details page, but it is possible to do so in the requests list.
Unless I'm misunderstanding the role, I believe this is a bug and the user should only be allowed to remove the requests **they** made.

**To Reproduce**
1. Request a movie as user A
2. Login as user B having "Manage Own Requests" role
3. Go to Requests page
4. User B can delete the request made by user A

**Expected behavior**
The API should return an error and the option should not be visible in the UI.

**Screenshots**
image

 - Version 4.4.0

sephrat:
- [x] API paths are secured (I removed "ManageOwnRequests" authorization from one unused API path that couldn't be secured by user)
- [x] "Delete request" option removed from the UI when the user isn't allowed

Fixes #4391

### DIFF
--- a/src/Ombi.Core/Engine/BaseMediaEngine.cs
+++ b/src/Ombi.Core/Engine/BaseMediaEngine.cs
@@ -78,23 +78,23 @@ namespace Ombi.Core.Engine
             return _dbTv;
         }
 
-        protected async Task<RequestEngineResult> CheckOwnRequests(BaseRequest request) {
+        protected async Task<RequestEngineResult> CheckCanManageRequest(BaseRequest request) {
             
-            var isRequestedBySameUser = ( await GetUser() ).Equals(request.RequestedUser);
+            // Admins can always manage requests
             var isAdmin = await IsInRole(OmbiRoles.PowerUser) || await IsInRole(OmbiRoles.Admin);
-            
-            if (!isRequestedBySameUser && !isAdmin)
-            {
-                return new RequestEngineResult
-                {
-                    Result = false,
-                    ErrorCode = ErrorCode.NoPermissions
-                };
-            }
+            if (isAdmin)
+                return new RequestEngineResult { Result = true };
+
+            // Users with 'ManageOwnRequests' can only manage their own requests
+            var isRequestedBySameUser = ( await GetUser() ).Equals(request.RequestedUser);
+            var canManageOwnRequests = await IsInRole(OmbiRoles.ManageOwnRequests);
+            if(canManageOwnRequests && isRequestedBySameUser)
+                return new RequestEngineResult { Result = true };
 
             return new RequestEngineResult
             {
-                Result = true,
+                Result = false,
+                ErrorCode = ErrorCode.NoPermissions
             };
         }
 
--- a/src/Ombi.Core/Engine/BaseMediaEngine.cs
+++ b/src/Ombi.Core/Engine/BaseMediaEngine.cs
@@ -78,6 +78,26 @@ namespace Ombi.Core.Engine
             return _dbTv;
         }
 
+        protected async Task<RequestEngineResult> CheckOwnRequests(BaseRequest request) {
+            
+            var isRequestedBySameUser = ( await GetUser() ).Equals(request.RequestedUser);
+            var isAdmin = await IsInRole(OmbiRoles.PowerUser) || await IsInRole(OmbiRoles.Admin);
+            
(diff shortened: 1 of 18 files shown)


src/Ombi.Core/Engine/BaseMediaEngine.cs
```diff
@@ -78,6 +78,26 @@ public int ResultLimit
...
         }
 
+        protected async Task<RequestEngineResult> CheckCanManageRequest(BaseRequest request) {
+            
+            // Admins can always manage requests
+            var isAdmin = await IsInRole(OmbiRoles.PowerUser) || await IsInRole(OmbiRoles.Admin);
+            if (isAdmin)
+                return new RequestEngineResult { Result = true };
+
+            // Users with 'ManageOwnRequests' can only manage their own requests
```
tidusjar, line 88: We could do this check first as it's potentially quicker than retrieving the user from the DB

something like 
```

var canManageOwnRequests = await IsInRole(OmbiRoles.ManageOwnRequests);
if (!canManageOwnRequests)
{
    return new RequestEngineResult
    {
        Result = false,
        ErrorCode = ErrorCode.NoPermissions
    };
}
// Users with 'ManageOwnRequests' can only manage their own requests
var isRequestedBySameUser = (await GetUser().Id).Equals(request.RequestedUser.Id);
if (isRequestedBySameUser)
{
    return new RequestEngineResult { Result = true };
}
```

src/Ombi.Core/Engine/BaseMediaEngine.cs
```diff
@@ -78,6 +78,26 @@ public int ResultLimit
...
+            var isRequestedBySameUser = ( await GetUser() ).Equals(request.RequestedUser);
```
tidusjar, line 89: `( await GetUser() ).Equals(request.RequestedUser);`
An object comparison might not be a good idea here as the object is coming from the request and the `GetUser` might differ slightly. So just comparing the `Id` property would be enough

src/Ombi.Core/Engine/BaseMediaEngine.cs
```diff
@@ -78,6 +78,26 @@ public int ResultLimit
...
+            var canManageOwnRequests = await IsInRole(OmbiRoles.ManageOwnRequests);
+            if(canManageOwnRequests && isRequestedBySameUser)
+                return new RequestEngineResult { Result = true };
```
tidusjar, line 92: Can we please add braces around the `if` statements

src/Ombi/ClientApp/src/app/requests-list/components/movies-grid/movies-grid.component.ts
```diff
@@ -46,10 +47,13 @@ export class MoviesGridComponent implements OnInit, AfterViewInit {
...
 
     constructor(private requestService: RequestServiceV2, private ref: ChangeDetectorRef,
-                private auth: AuthService, private storageService: StorageService,
-                private requestServiceV1: RequestService, private notification: NotificationService,
-                private translateService: TranslateService) {
+        private auth: AuthService, private storageService: StorageService,
+        private requestServiceV1: RequestService, private notification: NotificationService,
+        private translateService: TranslateService, private identity: IdentityService) {
 
+        identity.getUser().subscribe(u => {
```
tidusjar, line 54: I think using the `auth.service.ts` and getting the Username from the claim would be better than calling an API?

tidusjar (changes requested):

src/Ombi.Core/Engine/BaseMediaEngine.cs
sephrat: Makes sense, but FYI IsInRole also calls GetUser() so it will just prevent calling it twice for nothing. Maybe this method could be improved but it's out of scope of this PR.

src/Ombi.Core/Engine/BaseMediaEngine.cs
tidusjar: Don't think we did this change? 😀

sephrat: Oops, I did but then I undid it. Stay tuned.

tidusjar (changes requested):

PR #4398, opened 2021-11-10
Better URL encoding

### LINKED ISSUE #4065: Movies with reserved URI characters return incorrect results
**Describe the bug**
Query parameters in the Ombi API are not escaped which causes movies with reserved uri characters return incorrect results.

**To Reproduce**
1. Go to the discover section of the website
2. Search for the movie "In & Of Itself"
3. The movie (https://www.imdb.com/title/tt11916302/) appears nowhere in the results

**Expected behavior**
The movie should appear in the results, and if you try and query TheMovieDB api directly you will see it appears first.

**Desktop (please complete the following information):**
 - OS: Arch Linux / OS X

- Version: 3.0.5223
- Media Server: Plex
- Database Type: SQLite

**Additional context**
I've never touched C# before, but I believe escaping the query using the Uri.DataEscapeString function will prevent this from occurring.

sephrat:
Fixes #4065 

There were two issues around the same topic:

- Discover search query from frontend to backend was not properly encoded, so searching for "AC/DC" would break the routing (error 500). I couldn't think of any other routes that may also be affected.
- Query from backend to external APIs were not properly encoded, so searching for "&" would be interpreted as a new query parameter by the external service.

It's a big change but I can't find any regression so far, only better search results.

### DIFF
--- a/src/Ombi.Api/ApiHelper.cs
+++ b/src/Ombi.Api/ApiHelper.cs
@@ -66,6 +66,8 @@ namespace Ombi.Api
                 startingTag = builder.Query.Contains("?") ? "&" : "?";
             }
 
+            value = Uri.EscapeDataString(value);
+            
             builder.Query = hasQuery
                 ? $"{builder.Query}{startingTag}{parameter}={value}"
                 : $"{startingTag}{parameter}={value}";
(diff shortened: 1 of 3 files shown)


r6rob:
Thank you! Fix is working perfectly [Screenshot_20220402-233016_Ombi](https://user-images.githubusercontent.com/73448667/161385583-d67618ba-6975-46f7-9c7f-65f309070910.jpg)

PR #4457, opened 2022-01-05
Let users see the full anonymous list of requests added.

maartenheebink:
As requested by myself and https://features.ombi.io/suggestions/114556/ability-to-shows-requests-to-other-users-but-hide-requested-by


src/Ombi.Store/Repository/Requests/MovieRequestRepository.cs
```diff
@@ -46,12 +46,19 @@ public MovieRequests GetRequest(int theMovieDbId)
...
-            return Db.MovieRequests
+            var allRequests = Db.MovieRequests
                 .Include(x => x.RequestedUser)
                 .ThenInclude(x => x.NotificationUserIds)
-                .AsQueryable();
+                .ToList();
+
+            if (anonimize)
+            {
+                allRequests.ForEach(x => x.RequestedUser = null);
```
tidusjar, line 58: We are iterating over the list of `allRequests` here twice, it would be better just to iterate over it once and set both properties to `null` in a single loop

src/Ombi/ClientApp/src/app/requests-list/components/movies-grid/movies-grid.component.ts
```diff
@@ -102,7 +103,15 @@ export class MoviesGridComponent implements OnInit, AfterViewInit {
...
                     this.isLoadingResults = false;
                     this.resultsLength = data.total;
-
+                    if (data.collection.filter(x => x.requestedUserId != null).length > 0 &&
+                      data.collection.filter(x => x.requestedUser != null).length > 0) {
+                      this.anonimized = false;                      
+                    }
+                    if (this.anonimized) {
+                      this.displayedColumns.forEach((element, index) => {
+                        if (element == 'requestedUser.requestedBy') this.displayedColumns.splice(index, 1);
```
tidusjar, line 112:
```suggestion
                        if (element === 'requestedUser.requestedBy')
                        {
                            this.displayedColumns.splice(index, 1);
                        }
```

tidusjar:
initial changes look good, some failing unit tests :)

maartenheebink:
> initial changes look good, some failing unit tests :)

Yea, i was just tinkering with it yesterday. There are some pretty serious bugs still in it.

tidusjar:
Yeah no worries. I've not pulled the code down and tested it myself yet. But let me know when you think it's good to go and I'll have a crack at it 😀

src/Ombi/ClientApp/src/app/requests-list/components/movies-grid/movies-grid.component.ts
maartenheebink: JS/TS not the main language i code in and i always mess up strict equality :')

src/Ombi.Store/Repository/Requests/MovieRequestRepository.cs
maartenheebink: Isn't it so for LINQ that every chained method gets executed separately anyway, so executing the ForEach in the chain above would still result in double iteration? Not quite up to speed about the inner workings of LINQ though.

tidusjar: Not sure I'm following, my suggestion would be,

```
foreach (var request in allRequests)
{
    request.RequestedUser = null;
    request.RequestedUserId = null;
}

```

That would be one iteration over the collection.


Looking a bit more in-depth, we could end up with a performance issue here. Originally the dataset would have been an `IQueryable`, and as part of EF Core that would mean that the query would not be evaluated on the database until some sort of result set is called e.g. `.ToList()`, before that happens that majority of methods calling this did further filtering on the MovieRequests, for example, paging or availability filtering, which reduced the results returned from the database for a quicker and less intensive query.

But now we are grabbing every single movie request from the database every time, then filtering an in-memory list, rather than building the query. If an Ombi setup has a few thousand movie requests (which lots do) then we are going to introduce a slowdown as we are asking the database for every request every time.

The solution to this would be to loop over the requests and null out the users after the query has been executed (so after a `.ToList()` for example). That would be a better solution here.

maartenheebink: You're totally right, i didn't factor in the fact i'm building a query, executing it and then converting it back to a query to be executed again. Will have a look at this in the coming days!

src/Ombi.Store/Repository/Requests/MovieRequestRepository.cs
```diff
@@ -46,12 +46,18 @@ public MovieRequests GetRequest(int theMovieDbId)
...
-                .AsQueryable();
+            if (!anonimize)
+            {
+                return Db.MovieRequests
+                    .Include(x => x.RequestedUser)
+                    .ThenInclude(x => x.NotificationUserIds)
+                    .AsQueryable();
+            } else
+            {
+                return Db.MovieRequests.AsQueryable(); //This still populates the RequestedUser for the logged in user (or so it seems...)
```
tidusjar, line 59: Had a quick look at the docs, looks like this is because that EF Core has already potentially cached the data from another execution and it's using a shared database context.

It might be better to revert the changes to this `GetWithUser` method and instead create a new method that returns 

```
 return Db.MovieRequests.AsNoTracking().AsQueryable();
 ```

That will then disable the caching for that entity, but it also means that we cannot perform any updates to the entity without re-attaching to the context, so that's why I wouldn't do it as part o the `GetWithUser` method.


Instead maybe create a common method in the `MovieRequestEngine`, something like `GetRequests(bool anonymize)` and then either call `GetWithUser` or your new method depending on the value of `anonymize`

What do you think?

maartenheebink: Seems good, alternative option would be to pass a default parameter in `anonimize`, preferably false, as that would result in the same logic but without the split in the method but that's codestyling preference i guess?

tidusjar: Yeah that works for me too.

maartenheebink:
@tidusjar could you have a crack at it? Might be very close to completion i guess?

tidusjar:
Sure will check it out soon

tidusjar:
OK, some initial testing.

From a movie side it all looks good
From a TV show, it seems that we still bring back the user, User `c` Requested this show and I am logged in as user `b`
image

It's part of the `api/v1/Request/tv/{requestId}/child` API call

See here: https://github.com/maartenheebink/Ombi/blob/feature/see_requests/src/Ombi.Core/Engine/TvRequestEngine.cs#L614

maartenheebink:
I'll hop on this soon, in between some other things right now and couldn't/can't find the time

PR #4462, opened 2022-01-10
Send new request email notifications to power users

sephrat:
This provides the possibility to send "New Request" notifications to power users instead of the Ombi administrator.
On top of that, if Ombi has multiple admins, an email will be sent to all admins instead of the "Admin Email" set up in the notifications settings.

I tagged this as a draft because I'm not sure about the setting I added:
1. I'm not sure it should even be a setting since it makes more sense toggled on IMO. Why send the notification to a fraction of the population that is allowed to approve the request? 
2. This could be extended to mobile notifications (at least), but then that would make it a more global setting than an email one - probably a "New Request" notification template one?

### DIFF
--- a/src/Ombi.Notifications/Agents/LegacyMobileNotification.cs
+++ b/src/Ombi.Notifications/Agents/LegacyMobileNotification.cs
@@ -23,7 +23,7 @@ namespace Ombi.Notifications.Agents
         public LegacyMobileNotification(IOneSignalApi api, ISettingsService<MobileNotificationSettings> sn, ILogger<LegacyMobileNotification> log, INotificationTemplatesRepository r,
             IMovieRequestRepository m, ITvRequestRepository t, ISettingsService<CustomizationSettings> s, IRepository<NotificationUserId> notification,
             UserManager<OmbiUser> um, IRepository<RequestSubscription> sub, IMusicRequestRepository music, IRepository<Issues> issueRepository,
-            IRepository<UserNotificationPreferences> userPref) : base(sn, r, m, t, s, log, sub, music, userPref)
+            IRepository<UserNotificationPreferences> userPref) : base(sn, r, m, t, s, log, sub, music, userPref, um)
         {
             _api = api;
             _logger = log;
@@ -59,7 +59,7 @@ namespace Ombi.Notifications.Agents
             };
 
             // Get admin devices
-            var playerIds = await GetAdmins(NotificationType.NewRequest);
+            var playerIds = await GetAdmins();
             await Send(playerIds, notification, settings, model, true);
         }
 
@@ -77,7 +77,7 @@ namespace Ombi.Notifications.Agents
             };
 
             // Get admin devices
-            var playerIds = await GetAdmins(NotificationType.Issue);
+            var playerIds = await GetAdmins();
             await Send(playerIds, notification, settings, model);
         }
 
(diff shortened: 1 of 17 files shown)


tidusjar:
Yeah,

I mentioned a setting to prevent any issues and make it backwards compatible. But I agree it does make sense to be a default, if someone doesn't like it then we can alter that in the future.

Also regarding the other notification types, it should really apply to all (inc. mobile).

sephrat:
I've added the logic to mobile notifications, although I could not test this myself.
I could not find the logic for recipients determination for other notifications types such as Pushbullet, Discord, ... . How do you take into account users' tokens and how do you know in which case to send it to who?

tidusjar:
For the other agents, the notifications are published to the channels, so it's only really Email and Mobile that will go to individual users currently

src/Ombi.Notifications/BaseNotification.cs
```diff
@@ -210,6 +214,13 @@ protected UserNotificationPreferences GetUserPreference(string userId, Notificat
                 .FirstOrDefault(x => x.Agent == agent && x.UserId == userId);
         }
 
+        protected async Task<IEnumerable<OmbiUser>> GetPowerUsers()
```
tidusjar, line 217: This is not really getting power users, It's getting all users that have raised privilidges

src/Ombi.Notifications/Agents/LegacyMobileNotification.cs
```diff
@@ -241,15 +241,25 @@ protected override async Task Test(NotificationOptions model, MobileNotification
             await Send(playerIds, notification, settings, model);
         }
 
-        private async Task<List<string>> GetAdmins(NotificationType type)
+        private async Task<List<string>> GetAdmins()
         {
-            var adminUsers = (await _userManager.GetUsersInRoleAsync(OmbiRoles.Admin)).Select(x => x.Id).ToList();
+            return await UsersToPlayerIds(await _userManager.GetUsersInRoleAsync(OmbiRoles.Admin));
+        }
+        private async Task<List<string>> GetPowerUsersPlayerIds()
```
tidusjar, line 248: This is not used?

src/Ombi.Notifications/Agents/LegacyMobileNotification.cs
```diff
@@ -241,15 +241,25 @@ protected override async Task Test(NotificationOptions model, MobileNotification
             await Send(playerIds, notification, settings, model);
         }
 
-        private async Task<List<string>> GetAdmins(NotificationType type)
+        private async Task<List<string>> GetAdmins()
```
tidusjar, line 244: I suggest we just rename this to `GetNotificationRecipients`.

    private async Task<List<string>> GetNotificationRecipients() => await UsersToPlayerIds(await GetAdminAndPowerRecipients());

tidusjar (changes requested):

tidusjar:
We probably want to apply this to the following notification types:

* NewIssue
* IssueComment
* AddedToRequestQueue


I think we should just remove `GetAdmins()` from the MobileNotification.cs and always use `GetPrivilegedUsersPlayerIds`

sephrat:
Mmmh, I thought about it but tend to disagree on this. My interpretation of the difference between an admin and a power user is as follows:

- Admins are tech-savvy users, they're the ones who will know about the *arr apps, so if there's an issue with a media or an integration with the *arr, they're the ones who can sort this. 
- Power users are users who can also review requests, they're not necessarily tech-savvy. They don't have access to the settings page so they're not supposed to know how to sort out issues or integration errors.

Does that make sense?

tidusjar:
Fair, you got me there. I agree with your points... I guess that they couldn't resolve issues either in that case...

PR #4463, opened 2022-01-10
Media servers + newsletter refactoring

sephrat:
- [x] Merge common code between Plex, Emby and Jellyfin
  - [x] Repository
  - [x] Content
  - [x] Episode
- [x] Make use of this in NewsletterJob to remove duplicate code
- [x] Save absolute URL for Plex content in DB (be consistent with Emby and Jellyfin)


tidusjar:
Oh man this has made my day! This is something I've wanted to do for a long time but never had the time to do it!

sephrat:
I just hit a major milestone. I can now focus on actually refactoring the newsletter.
There's probably more to do with abstracting the media servers: there are still some unnecessary differences. However the most needed parts are now reworked and should allow for a clean refactoring.
I'll get back to it later, this was exhausting 😄

src/Ombi.Core/Rule/Rules/Search/AvailabilityRuleHelper.cs
```diff
@@ -79,66 +79,6 @@ public static void CheckForUnairedEpisodes(SearchTvShowViewModel search)
                 log.LogError(e, "Exception thrown when attempting to check if something is available");
             }
 
-            if (epExists != null)
```
tidusjar, line 82: why was all of this removed?

src/Ombi.Schedule/Jobs/Ombi/NewsletterJob.cs
```diff
@@ -124,18 +124,22 @@ await _notification.Clients.Clients(NotificationHub.AdminConnectionIds)
 
 
                 var customization = await _customizationSettings.GetSettingsAsync();
-                // Get the Content
-                var plexContent = _plex.GetAll().Include(x => x.Episodes).AsNoTracking();
-                var embyContent = _emby.GetAll().Include(x => x.Episodes).AsNoTracking();
-                var jellyfinContent = _jellyfin.GetAll().Include(x => x.Episodes).AsNoTracking();
-                var lidarrContent = _lidarrAlbumRepository.GetAll().AsNoTracking().ToList().Where(x => x.FullyAvailable);
+                // MOVIES
```
tidusjar, line 127: It's not really just movies, it's Movies + Shows and Episodes (due to the `.include`)

src/Ombi.Schedule/Jobs/Ombi/NewsletterJob.cs
```diff
@@ -375,21 +377,32 @@ await _notification.Clients.Clients(NotificationHub.AdminConnectionIds)
...
-            addedEmbyMoviesLogIds = embyParent != null && embyParent.Any() ? (embyParent?.Select(x => x.ContentId)?.ToHashSet() ?? new HashSet<int>()) : new HashSet<int>();
-
-            var jellyFinParent = addedLog.Where(x => x.Type == RecentlyAddedType.Jellyfin && x.ContentType == ContentType.Parent);
-            addedJellyfinMoviesLogIds = jellyFinParent != null && jellyFinParent.Any() ? (jellyFinParent?.Select(x => x.ContentId)?.ToHashSet() ?? new HashSet<int>()) : new HashSet<int>();
-
             var lidarrParent = addedLog.Where(x => x.Type == RecentlyAddedType.Lidarr && x.ContentType == ContentType.Album);
             addedAlbumLogIds = lidarrParent != null && lidarrParent.Any() ? (lidarrParent?.Select(x => x.AlbumId)?.ToHashSet() ?? new HashSet<string>()) : new HashSet<string>();
         }
 
+        private async Task<HashSet<IMediaServerContent>> getMoviesContent(IQueryable<IMediaServerContent> content, RecentlyAddedType recentlyAddedType)
```
tidusjar, line 386:
```suggestion
        private async Task<HashSet<IMediaServerContent>> GetMoviesContent(IQueryable<IMediaServerContent> content, RecentlyAddedType recentlyAddedType)
```

src/Ombi.Core/Rule/Rules/Search/AvailabilityRuleHelper.cs
sephrat: git diff is a bit stupid, the last bits of this method are still there after the part I removed (l.142/82).
If you were asking about the whole section being removed: the Plex/Jellyfin/Emby methods were merged into a single one which takes IMediaServerEpisode as an input.

src/Ombi.Schedule/Jobs/Ombi/NewsletterJob.cs
sephrat: You're right. This part should change a lot by the end of this PR anyway.

sephrat:
I think it's in a good shape now. I've reassessed the scope of this PR, see my edited OP.
There's still some work to do in removing duplicate code between Emby/Jellyfin/Plex in some other places, but the foundations are there to allow this in a future PR.
Once this gets merged, I'll work on the actual PR I initially intended to work on for the newsletter.

tidusjar:
Looks like we have some unit test failures :( I'm planning to do some testing around this soon

sephrat:
Oops, didn't see that one. Pretty sure it works in productive mode though, I've simplified the URL part when merging Plex with Emby and Jellyfin: they're calculated once upon sync. The tests probably only need to be rewritten to check the URL on sync, not on availability check.

tidusjar:
> Oops, didn't see that one. Pretty sure it works in productive mode though, I've simplified the URL part when merging Plex with Emby and Jellyfin: they're calculated once upon sync. The tests probably only need to be rewritten to check the URL on sync, not on availability check. I'll look into it.

That makes sense for new syncs, but what about existing synced data?

sephrat:
Fair point. It can easily be fixed by forcing a full resync, but I guess it'd be better not to break this. I'll see what I can do. For the record, this is only relevant for Plex - Emby and Jellyfin syncs already store the full URL in the database.

tidusjar:
If there's nothing simple, then we can just recommend a clear and full resync. It would be a very temporary issue.

sephrat:
I've removed the tests since I could not find a simple way to migrate them into sync tests (sorry), and it's not AvaibilityRule's responsibility anymore.
I managed to find a simple fix for legacy Plex content URL.
Should be all good now.

tidusjar:
I've been testing this and awesome work! Almost ready to merge it 🤞

PR #4468, opened 2022-01-13
A more reliable Emby and Jellyfin sync

sephrat:
There is an offset error in Emby sync: Emby API offset is 0-based and Ombi syncs from 0-99 then from 101-200.
This means the 100th movie, the 100th TV episode of a show, and the 100th show would not get synced. Now if that's not a bug that made me think I was crazy...!


sephrat:
FYI Jellyfin has the same logic but I don't know if the API behaves the same.

tidusjar:
Haha nice find! Yes the api for Jellyfin behaves the same. Would you mind making the same chances over there?

sephrat:
Done!

Issue #3921, opened 2020-12-08
V4 - Application Name in Top Left Corner is not Linkified


aman88:
**Describe the bug**
Clicking on the Application Name in the top left corner should bring you back to the application home page. Currently, in V4, this is not a linkified item.

**To Reproduce**
1. Navigate to any Ombi page aside from the home page
2. Click on the Application Name in the top left corner
3. It's not a link, so nothing happens.

**Expected behavior**
Clicking the Application Name should redirect you to the application home page.

**Ombi Version (please complete the following information):**
 - Version 4.0.739.0
- Media Server - Plex
- Database Type: SQLite

tidusjar:
This has been fixed in the develop branch! 🎉 🎈

PR #4477, opened 2022-01-23
Add link on logo text

LMaxence:
I just went across [this issue](https://github.com/Ombi-app/Ombi/issues/3921) that seemed to be close to a one-liner to satisfy

### DIFF
--- a/src/Ombi/ClientApp/src/app/my-nav/my-nav.component.html
+++ b/src/Ombi/ClientApp/src/app/my-nav/my-nav.component.html
@@ -3,7 +3,12 @@
         [attr.role]="(isHandset$ | async) ? 'dialog' : 'navigation'" [mode]="(isHandset$ | async) ? 'over' : 'side'"
         [opened]="!(isHandset$ | async)">
 
-        <mat-toolbar class="application-name" id="nav-applicationName"><a class="application-name-link" [href]="applicationUrl">{{applicationName}}</a></mat-toolbar>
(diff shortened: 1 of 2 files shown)


tidusjar:
Nice! Looks good

src/Ombi/ClientApp/src/app/my-nav/my-nav.component.html
```diff
@@ -3,7 +3,7 @@
         [attr.role]="(isHandset$ | async) ? 'dialog' : 'navigation'" [mode]="(isHandset$ | async) ? 'over' : 'side'"
         [opened]="!(isHandset$ | async)">
 
-        <mat-toolbar class="application-name" id="nav-applicationName">{{applicationName}}</mat-toolbar>
+        <mat-toolbar class="application-name" id="nav-applicationName"><a class="application-name-link" [href]="applicationUrl">{{applicationName}}</a></mat-toolbar>
```
tidusjar, line 6: One thing I just thought about, `applicationUrl` could be null.

It might be worth applying the `<a href` when the applicationUrl is not null

tidusjar (changes requested):

tidusjar (approved):

tidusjar:
Some CI fails because it's a PR from a non member so it doesn't have permission to run. Nothing to worry about. I'll get it merged in soon! Thanks :)

PR #4483, opened 2022-01-29
fix swagger api key auth

cqxmzz:
the api auth for swagger page has been broken
http://localhost:5000/swagger/index.html

and according to swagger doc
https://swagger.io/docs/specification/authentication/api-keys/
we are setting it up wrong

WX20220128-162208

local test:
able to get response
WX20220128-161724

swagger json config looks good
WX20220128-162255


tidusjar:
Nice looks good!

PR #4485, opened 2022-02-01
Newsletter localization

sephrat:
This will allow some strings in the newsletter to be translated through the use of .resx files. These files are supported by Crowdin and could be added there to allow for global localization. I've already translated the FR part for testing purposes and included it in this PR.

For now the newsletter will use the global language settings, it won't take into account the user's individual preferences. 
It will use English as default if there is no resource file present for the set language, so it should be safe to deploy before integrating it into Crowdin and have the generated files for every language.

I hope I did right by the guidelines with the I18n packaging, this is fairly new to me. I'm open to feedback.


sephrat:
I worked on this before the refactoring was merged to develop, sorry for the mess in the commit list. The relevant commits are near the bottom (from 45c5c1e).

Issue #4444, opened 2021-12-17
Newsletter sent with a "New TV" block but no item below


sephrat:
**Describe the bug**
Just received my weekly newsletter with a "New TV" block without any item below.
I think there are two inherent issues:

- The block shouldn't display at all instead of showing without any item.
- One item should have popped in the list. The lead I have so far: it was added to Emby a few hours before the newsletter job, but is not referenced as "Available" in Ombi yet. Maybe there's an issue with the newsletter when Emby and Ombi are out of sync?
Edit: actually there's a second item that should have popped and is properly synced in Ombi, so that's not it. Will try to find more useful info.

**Screenshots**
![image](https://user-images.githubusercontent.com/34862846/146543406-a6b3f62c-c890-4708-b996-ac77f1a68d83.png)

**Logs (Logs directory where Ombi is located)**
```
2021-12-17 12:00:00.001 +01:00 [Debug] Calling Execute on job System.INewsletterJob
2021-12-17 12:00:00.005 +01:00 [Debug] Trigger instruction : NoInstruction
2021-12-17 12:00:17.923 +01:00 [Error] StatusCode: NotFound, Reason: Not Found, RequestUri: http://api.themoviedb.org/3/find/?api_key=b8eabaf5608b88d0298aa189dd90bf00&external_source=imdb_id
2021-12-17 12:00:19.432 +01:00 [Error] StatusCode: NotFound, Reason: Not Found, RequestUri: http://api.tvmaze.com/lookup/shows?thetvdb=82890
2021-12-17 12:00:19.480 +01:00 [Error] StatusCode: NotFound, Reason: Not Found, RequestUri: http://api.tvmaze.com/lookup/shows?thetvdb=412366
```

**Ombi Version (please complete the following information):**
 - Version 4.7.8
- Media Server : Emby

sephrat:
Ah, I think I see it. The newsletter generates the "New TV" block when there are new episodes but the individual episodes won't show if no info could be found from TVMaze API. 
https://github.com/Ombi-app/Ombi/blob/5e140ab6183b887a7665f5e870eb0bd05d487ace/src/Ombi.Schedule/Jobs/Ombi/NewsletterJob.cs#L1069-L1073
We need to generate the "TV block" only when we know for sure some content will be generated after.
On top of that, shouldn't we move from TVMaze to TMDB to be consistent with the rest of the app?

tidusjar:
Looks like that's an area we missed migrating from from TvMaze

PR #4486, opened 2022-02-01
Drop TVMaze from Newsletter

sephrat:
Removed some leftover from TVMaze API uses in the newsletter.
Partial fix for #4444
This should hopefully allow for more TV shows to be sent in the newsletter.

Also added some logs to better understand when an item won't be sent to the newsletter, in the hope to come up with a better fix for #4444.


tidusjar:
This looks good, all ready to merge?

sephrat:
Yes I've tested it already.

PR #4493, opened 2022-02-04
Add sanity checks upon media server sync

### LINKED ISSUE #4472: Media content may be improperly imported into Ombi
**Describe the bug**
If Ombi fetches content from a media server when it isn't done fetching the metadata, it will save the content in the database in an incorrect state:
image

Upon the next sync, Ombi considers the content as already added so it won't try to update its information. This ends up in content that can never be marked as available unless you manually run a "Clear data and resync" in Ombi settings.

This causes other issues in another context: if the media was first scanned incorrectly by the media server and you manually correct the identification, Ombi will keep the incorrect state in the database.

**To Reproduce**
1. Add new content to media server
2. Immediately run a sync (after it has detected new content but before the metadata are recognized).
3. Wait for the metadata to be added in the media server and run a new sync
4. The media is not available in Ombi

**Expected behavior**
Ombi should update the media information if it detects a difference.

**Proposed solution**
- Do not add content to the database when its metadata is empty (no external IDs, no episode/season number)
- Update content when there's a difference between the media server state and Ombi internal state (an ID has changed, an episode number was changed, etc. )

 - Version 4.9.1
- Media Server Emby

Trying to get some feedback before working on a PR.

sephrat:
Fixes #4472

### DIFF
--- a/src/Ombi.Api.Emby/Models/Media/EmbyProviderids.cs
+++ b/src/Ombi.Api.Emby/Models/Media/EmbyProviderids.cs
@@ -1,27 +1,11 @@
-﻿namespace Ombi.Api.Emby.Models.Movie
-{
-    public class EmbyProviderids
-    {
-        public string Tmdb { get; set; }
-        public string Imdb { get; set; }
-        public string TmdbCollection { get; set; }
+﻿using Ombi.Api.MediaServer.Models;
 
-        public string Tvdb { get; set; }
+namespace Ombi.Api.Emby.Models.Movie
+{
+    public class EmbyProviderids: BaseProviderids
+    {
+        public string TmdbCollection { get; set; }
         public string Zap2It { get; set; }
         public string TvRage { get; set; }
-        
-        public bool Any()
-        {
-            if (string.IsNullOrEmpty(Imdb)
-               && string.IsNullOrEmpty(Tmdb)
-               && string.IsNullOrEmpty(Tvdb))
-            {
-                return true;
-            }
-            else
-            {
-                return false;
-            }
-        }
     }
 }
--- a/src/Ombi.Api.Emby/Models/Media/EmbyProviderids.cs
+++ b/src/Ombi.Api.Emby/Models/Media/EmbyProviderids.cs
@@ -9,5 +9,19 @@
         public string Tvdb { get; set; }
         public string Zap2It { get; set; }
         public string TvRage { get; set; }
+        
+        public bool Any()
+        {
+            if (string.IsNullOrEmpty(Imdb)
+               && string.IsNullOrEmpty(Tmdb)
+               && string.IsNullOrEmpty(Tvdb))
+            {
+                return true;
+            }
+            else
+            {
+                return false;
+            }
+        }
     }
 }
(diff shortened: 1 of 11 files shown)


sephrat:
For the record, I do not check season number because Specials are usually referenced as Season 0.

src/Ombi.Api.Emby/Models/Media/EmbyProviderids.cs
```diff
@@ -9,5 +9,19 @@ public class EmbyProviderids
         public string Tvdb { get; set; }
         public string Zap2It { get; set; }
         public string TvRage { get; set; }
+        
+        public bool Any()
```
tidusjar, line 13: This correct? Any() will return true if they are all `NullOrEmpty`?

Maybe create a `BaseProviderIds` class?

```
public class BaseProviderIds
{
    public string Tmdb { get; set; }
    public string Imdb { get; set; }
    public string Tvdb { get; set; }

    public bool Any() => 
        !string.IsNullOrEmpty(Imdb) || !string.IsNullOrEmpty(Tmdb) || !string.IsNullOrEmpty(Tvdb);
}
```

src/Ombi.Api.Jellyfin/Models/Media/JellyfinProviderids.cs
```diff
@@ -9,5 +9,19 @@ public class JellyfinProviderids
         public string Tvdb { get; set; }
         public string Zap2It { get; set; }
         public string TvRage { get; set; }
+        
+        public bool Any()
```
tidusjar, line 13: You can use suggested base class here

src/Ombi.Schedule/Jobs/Emby/EmbyContentSync.cs
```diff
@@ -249,6 +249,13 @@ private async Task ProcessMovies(EmbyMovie movieInfo, ICollection<EmbyContent> c
             var alreadyGoingToAdd = content.Any(x => x.EmbyId == movieInfo.Id);
             if (existingMovie == null && !alreadyGoingToAdd)
             {
+
+                if (string.IsNullOrEmpty(movieInfo.ProviderIds?.Imdb)
```
tidusjar, line 253: Why not use the new `Any()`?

src/Ombi.Schedule/Jobs/Jellyfin/JellyfinContentSync.cs
```diff
@@ -217,6 +217,12 @@ private async Task ProcessMovies(JellyfinMovie movieInfo, ICollection<JellyfinCo
             var alreadyGoingToAdd = content.Any(x => x.JellyfinId == movieInfo.Id);
             if (existingMovie == null && !alreadyGoingToAdd)
             {
+                if (string.IsNullOrEmpty(movieInfo.ProviderIds?.Imdb)
```
tidusjar, line 220: Use `Any()`?

tidusjar (changes requested):
Looks good, just a few minor things :)

src/Ombi.Api.Emby/Models/Media/EmbyProviderids.cs
sephrat: Oh dear, don't know where my head was when I did that (and not using this method). Sorry for the half-baked PR.
Also thought about a base class but was not sure how to package it across Emby and Jellyfin, I'll mull it over. Bare with me.

sephrat:
Should be much better now.

PR #4510, opened 2022-02-15
SSO with Header Authentication

fservida:
I've a reverse proxy in place, and wanted for some time to integrate user management with SSO.
I've looked at different options and the easiest and possibly most versatile one was to implement header authentication, so that people can authenticate however they see fit on a reverse proxy and just set a header variable to the username to login.

This PR solves this problem by adding a setting in the Authentication page to enable user authentication via header variable, and lets the user specify a variable to use.
If enabled classic authentication is completely bypassed **if the user is found.** 
Needless to say that users have to be created somehow beforehand, either locally, by importing plex/emby.. or possibly in the future via LDAP if the other PR are merged.

If it is enabled, but the header is unset or the user is not known it reverts to the classic authentication.
A side case, but common to any SSO without SLO is that logout does not work anymore as the user is immediately logged in again.

Linked to https://feathub.com/tidusjar/Ombi/+460 and https://features.ombi.io/suggestions/117140/sso-oauthsaml

I admit it's kinda monkey patched in there, although I tried to handle every case gracefully, so I welcome code review from experienced contributors, let me know if something can be improved.

image

### DIFF
--- a/src/Ombi.Settings/Settings/Models/AuthenticationSettings.cs
+++ b/src/Ombi.Settings/Settings/Models/AuthenticationSettings.cs
@@ -13,5 +13,8 @@ namespace Ombi.Settings.Settings.Models
         public bool RequireNonAlphanumeric { get; set; }
         public bool RequireUppercase { get; set; }
         public bool EnableOAuth { get; set; } // Plex OAuth
+        public bool EnableHeaderAuth { get; set; } // Header SSO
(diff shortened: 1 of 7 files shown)


tidusjar:
Thanks. I’ll take a look and let you know what I think!

tidusjar:
Code looks good and seems to be a sensible approach

tidusjar:
No should be good. I just want to pull the code down and test it before I merge it in!

PR #4519, opened 2022-02-24
Remove generic admin email in favor of admins' email

sephrat:
This removes the generic admin email setting.
Instead we use the email addresses set on the users' profile.
Allows for notifications to many recipients in case of multiple admins.
Email testing now sends the test email to the currently logged in user.

### DIFF
--- a/src/Ombi.Notifications/Agents/EmailNotification.cs
+++ b/src/Ombi.Notifications/Agents/EmailNotification.cs
@@ -47,7 +47,7 @@ namespace Ombi.Notifications.Agents
                     return false;
                 }
             }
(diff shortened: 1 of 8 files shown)


src/Ombi.Notifications/Agents/EmailNotification.cs
```diff
@@ -73,26 +73,6 @@ private async Task<NotificationMessage> LoadTemplate(NotificationType type, Noti
...
-                {
-                    // Send to admin
-                    message.To = settings.AdminEmail;
-                }
-            }
-            else
-            {
-                // Send to admin
-                message.To = settings.AdminEmail;
-            }
```
sephrat, line 95: This seems to be dead/redundant code, my tests went fine when removing this. 
See possibly redundant code: https://github.com/Ombi-app/Ombi/pull/4519/files#diff-69436d6a4ec4479e1f5a265b7427ed6fe412d1a060eb8e6275e8574b89cf2c91R135

PR #4525, opened 2022-02-28
translation(cs): adds czech translation

DavidReinberger:


sephrat:
Please use Crowdin to submit new translations: https://docs.ombi.app/development/translations/

PR #4533, opened 2022-03-03
Fix some Radarr settings no longer initialized in settings page

sephrat:
This was lost in the 4K refactoring.


tidusjar:
Nice find!

PR #4576, opened 2022-04-06
Plex watchlist requests

tidusjar:
note: this will only work for users that are Plex Users. For this to work straight away the Plex User needs to login to Ombi (If they are already logged in, logout and back in).

It sync's currently on a daily schedule.

### DIFF
--- a/src/Ombi.Schedule/Jobs/Plex/PlexWatchlistImport.cs
+++ b/src/Ombi.Schedule/Jobs/Plex/PlexWatchlistImport.cs
@@ -1,0 +1,22 @@
+﻿using Quartz;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Ombi.Schedule.Jobs.Plex
+{
+    public class PlexWatchlistImport : IPlexWatchlistImport
+    {
+        public void Dispose()
+        {
+            throw new NotImplementedException();
+        }
+
+        public Task Execute(IJobExecutionContext context)
+        {
+            throw new NotImplementedException();
+        }
+    }
+}
--- a/src/Ombi.Schedule/Jobs/Plex/PlexWatchlistImport.cs
+++ b/src/Ombi.Schedule/Jobs/Plex/PlexWatchlistImport.cs
@@ -12,6 +12,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Ombi.Schedule.Jobs.Plex
@@ -38,7 +39,6 @@ namespace Ombi.Schedule.Jobs.Plex
 
         public async Task Execute(IJobExecutionContext context)
         {
-            var token = "-DpQi6mzq2QMakYgFr2g"; // !!!!!!!!!!!!!!!!!!!! TODO REMOVE !!!!!!!!!!!!!!!!!!!!!!!
 
             var settings = await _settings.GetSettingsAsync();
             if (!settings.Enable)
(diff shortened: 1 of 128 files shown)


src/Ombi.Schedule/Jobs/Plex/PlexWatchlistImport.cs
```diff
@@ -0,0 +1,124 @@
...
+            _settings = settings;
+            _ombiUserManager = ombiUserManager;
+            _movieRequestRepository = movieRequestRepository;
+            _tvRequestRepository = tvRequestRepository;
+            _movieRequestEngine = movieRequestEngine;
+        }
+
+        public async Task Execute(IJobExecutionContext context)
+        {
+            var token = "-DpQi6mzq2QMakYgFr2g"; // !!!!!!!!!!!!!!!!!!!! TODO REMOVE !!!!!!!!!!!!!!!!!!!!!!!
```
tidusjar, line 41: nevermind...

malicioustoker:
What exactly does this do? 🤔

Firetop:
> What exactly does this do? 🤔

I believe it's suppose to Request in Ombi whatever you added to your Plex Watchlist.

Sadly I've tested this extensively and it does not seem not work for me. @sephrat and @tidusjar 

I turned up the Cron rate to every 5 minutes; restarted Ombi multiple times; signed out and signed in multiple times with hard F5 refreshes between each logout and login; manually ran the Watchlist Importer from the Plex settings page and it also does nothing for me.

Checking the logs shows no results either when running the manual Watchlist Importer; 

image

I was hoping for some kind of error to provide but can't even find one... Feels half baked rn since the idea itself is nice but in practice it needs more work before being pushed to production it seems; or more documentation as to what is actually required for this to function correctly; as the documentation of "enabling the feature and logging out and back in" does not make it work instantly for me anyways.

V4.16.2 using Sqlite on Windows 10 21H2

tidusjar:
I haven't yet wrote the documentation for this feature.

Thanks for letting me know it's not working as expected @Firetop.

What's likely happened is that the code to match the Ombi internal user with the Plex user is using an incorrect matching ID. Which means this will work fine for new users until that ID rotates. This is difficult as Plex don't actually document any of their API's so it's difficult to know.

> I was hoping for some kind of error to provide but can't even find one...
Because the Id's do not match, it doesn't know that there's a user with a  watchlist, in that case there is no error.

> it needs more work before being pushed to production 
It's not in production, it's in the develop branch. The aim of the develop branch is to find any issues like this before merging to master. I am one dev, there are hundreds of different configurations that people may use with ombi and external applications so it's almost impossible for me to test it all.

> more documentation as to what is actually required for this to function correctly
Yeah I spent quite a few hours coding this feature, was planning to do the documentation soon, I got pretty sleepy yesterday so skipped on that part as I didn't think it would matter if the docs came a bit later.

Thanks for the feedback.

tidusjar:
Turns out that I missed a code path, it's fixed in the next release v4.16.3

Firetop:
> I haven't yet wrote the documentation for this feature.
> 
> Thanks for letting me know it's not working as expected @Firetop.
> 
> What's likely happened is that the code to match the Ombi internal user with the Plex user is using an incorrect matching ID. Which means this will work fine for new users until that ID rotates. This is difficult as Plex don't actually document any of their API's so it's difficult to know.
> 
> > I was hoping for some kind of error to provide but can't even find one...
> 
> Because the Id's do not match, it doesn't know that there's a user with a watchlist, in that case there is no error.
> 
> > it needs more work before being pushed to production
> 
> It's not in production, it's in the develop branch. The aim of the develop branch is to find any issues like this before merging to master. I am one dev, there are hundreds of different configurations that people may use with ombi and external applications so it's almost impossible for me to test it all.

No I agree, that's what I was trying to say, right now it felt very much in its infancy, with a little more work the idea is great and would be ready for a push the main branch but I was definitely not trying to put you on the defensive, actually more like trying to back you up that with a bit more with this would be ready for production.

Thanks for digging into it so quickly as always, as for documentation that can take place when it goes to production, I was just saying if there was a specific feature in Plex we need to have enabled it should be documented because just logging out and back in again wasn't cutting it for me anyways...

I'll test the next release! Thanks again!

tidusjar:
@Firetop let me know how you get on!

Firetop:
> @Firetop let me know how you get on!

Was it really that easy? I just updated and tested and my previous whatlist items instantly started coming through (Since I had set the cron time super fast lol)

Fantastic! I honestly thought maybe I had something disabled that needed to be enabled since there was no errors I could see in the logs; but your explaination of the matching IDs makes complete sense.

Happy to know this was an easy fix for you and not something requiring many more hours of work lol!

Thanks again!
Joel S.

PR #4585, opened 2022-04-08
Fix subscribe button for movies

sephrat:
- Deprecate 'subscribed' field from search service which could be incorrect because of caching
- Fixes 'subscribed' field from request service
- Fixes 'Subscribe' button not being displayed for non-4K setups
- Allows admins to subscribe too

### DIFF
--- a/src/Ombi.Core/Engine/MovieRequestEngine.cs
+++ b/src/Ombi.Core/Engine/MovieRequestEngine.cs
@@ -251,7 +251,7 @@ namespace Ombi.Core.Engine
             var requests = await (OrderMovies(allRequests, orderFilter.OrderType)).Skip(position).Take(count)
                 .ToListAsync();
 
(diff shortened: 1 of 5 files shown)


tidusjar:
Now we have moved it, does the UI need to now set the properties from the request rather than the search object?

sephrat:
> Now we have moved it, does the UI need to now set the properties from the request rather than the search object?

It's been done here: https://github.com/Ombi-app/Ombi/commit/dd7219ebf9c2ca31acb89f034d0ae86b2c2d6894
Or did you mean something else?

tidusjar:
I missed that. That does make sense. My only issue is that if anyone is depending on that field in the api it’s not going to work. An example of this would be the mobile app. I suggest we put the v1 back for a bit and then later remove it. That way we can stage the rollout.

sephrat:
Sounds reasonable. Done.

Issue #4588, opened 2022-04-09
Ombi won't monitor unaired episodes of a season in Sonarr


sephrat:
**Describe the bug**
Requesting for a season in Ombi which contains aired and unaired episodes will result in only the aired episodes being monitored.
Individually requesting for the same episodes will work. It seems to happen only if the full season is requested. 

**To Reproduce**
Test sample:
/details/tv/52814
![image](https://user-images.githubusercontent.com/34862846/162578994-f577c045-b850-4ec9-a194-6b1c0e4bffb6.png)
Results in the following:
![image](https://user-images.githubusercontent.com/34862846/162578998-f4ea03da-1adf-42fa-8785-ff0f63ce7bbb.png)

**Analysis**
This seems to be specific to the "Seasonpass" call. That API method is not documented so it's hard to figure out what's wrong. Looking at Sonarr code, it looks like we're branching to legacy V2 API because we're not using the new monitoring option "Monitor": https://github.com/Sonarr/Sonarr/blob/0a2b109a3fe101e260b623d0768240ef8b7a47ae/src/NzbDrone.Core/Tv/EpisodeMonitoredService.cs#L37
Given it's V2, I assume it's deprecated and we won't get support from them.
We may want to look into the new features V3 API offers, and/or revamping our API calls to Sonarr.

**Side note**
While debugging this, I noticed that if you request for a full season + a few episodes of an other season, Ombi will monitor the two full seasons. Not your typical request I must say, but if we happen to revamp the code, we may want to look into it.

PR #4589, opened 2022-04-11
An attempt at simplifying Sonarr sync

sephrat:
I say "attempt" because this has a few drawbacks that I'll try to explain.

**Current behaviour:**

1. Ombi would detect requests for a full season and would monitor the full season in Sonarr with a simple API call.
    1. It's quite simple and optimized because it avoids unecessary individual API calls
    2. This has some unwanted side effects:
       1. see #4588
        2. Monitoring a full season will result in future episodes being monitored, which is not what the request should allow. I'm talking about episodes that did not exist in tmdb/tvdb when processing the request, but would be added later on. Typical case is seasons spanning over 2 years like the last season of Vikings.
2. When requesting for a few episodes of a season, it would monitor the whole season and then unmonitor episodes that were not requested and not already monitored.
    -  This is quite a complex logic, although it works just fine.

**New behaviour developed in this PR:**

- Ombi will monitor individual episodes, regardless of a full season or not.
  - This makes 100% sure every episode in the request is monitored - no more, no less. 
  - It can generate a high number of API calls which could slow down the syncing process for large requests on low-end servers (not benchmarked, just a feeling).

**Tests**
Requesting for 3 full seasons of a show + a few episodes of a season:
image
image
It may seem off that the season is not flagged as monitored despite all episodes being monitored, but this actually makes sense from a Sonarr perspective and is expected behaviour. The episodes do show up in the "Missing" tab.

**Additional information**
I would recommend dropping support for V2 with this PR for the following reasons:
- This is a major change that I haven't tested on V2 at all
- Sonarr V2 has been EOL for more than a year now, time to upgrade!

Fixes https://github.com/Ombi-app/Ombi/issues/4588

### DIFF
--- a/src/Ombi.Core/Senders/TvSender.cs
+++ b/src/Ombi.Core/Senders/TvSender.cs
@@ -343,7 +343,6 @@ namespace Ombi.Core.Senders
                 await Task.Delay(500);
             }
 
-            var seriesChanges = false;
 
             foreach (var season in model.SeasonRequests)
             {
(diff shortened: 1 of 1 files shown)


tidusjar:
I need to look at this more in-depth. But the problem with the season not being monitored is that if you do a search for that series, nothing will be found as sonarr skips unmonitored seasons

sephrat:
Mmmh you're right. Quite curious since the search will work from the Missing tab. I'll think it through.

sephrat:
>  the problem with the season not being monitored is that if you do a search for that series, nothing will be found as sonarr skips unmonitored seasons

Just to be sure we're talking about the same side-effects of having the season not monitored: 

1. Searching from the series' page in Sonarr with the global "Search monitored" button will indeed not trigger a search of the individually monitored episodes. This is a regression.
2. Searching from the series' page in Sonarr with the season search button or the episode search button works fine. So does manual searching from the Wanted > Missing tab in Sonarr.
3. Manual searching through the API that we use in Ombi works fine, be it either a season search or an episode search.
4. I'm still trying to figure out if auto-search in Sonarr works fine (when Sonarr automatically picks up newly released episodes through RSS sync). This not working would be a no-go for this PR in my opinion. I'll come back later with the result of my tests.

My personal opinion would be that side-effect n°1 is quite mitigated since in an automated setup, manual searches through Sonarr are an edge-case. In any case, it's a matter of pros and cons against the fact that monitoring the full season leads in future episodes being monitored despite not being part of the Ombi request. I'll let you be the judge of that. 😃

tidusjar:
> monitoring the full season leads in future episodes being monitored despite not being part of the Ombi request.

I think that’s fine. 

I think that what you will find with the RSS feed is that if the series or season is not monitored it will not get picked up. I’ll ask the devs to confirm

sephrat:
Thanks for looking into it. I'll come back with another approach.

tidusjar:
Thank you for looking into it in the first place!

Issue #4592, opened 2022-04-12
Video requests not being properly marked as available when only 4k quality is available


xFaultx:
**Describe the bug**
When a video that was requested is available in 4k, it is not being properly marked as available, showing it as 'Processing Request' on the request page. It should be showing as 'Available'.

**To Reproduce**
1. User requests video (note the 4kRequest feature is not turned on, but this has happened with both it on and off)
2. 4k version of the film is available on Plex server
3. User will see it marked as available on on the videos page, but will not get the 'available email'
4. Under requests, the film is still listed as 'Pending Approval' or 'Processing Request' even though the 4k version is available, both from a user and admin perspective.

**Expected behavior**
If a 4k version is available of the video, it should respond to user requests saying the file is available, regardless of 4krequests or not.

**Screenshots**
https://imgur.com/a/OfzGM73

**Desktop (please complete the following information):**
 - OS: Windows

**Ombi Version (please complete the following information):**
 - Version - 4.16.6
- Media Server - Plex
- Database Type: SQLite

**Additional context**
Poste about this on Reddit, was told to post the issue here

tidusjar:
Thanks! Do you mind sending me your Ombiexternal and Ombi.db files to tidusjar@gmail.com

PR #4594, opened 2022-04-13
fix(availability): Fixed an issue where we wouldn't mark a available 4k movie as available (when 4K request feature is disabled)

tidusjar:


tidusjar:
#4592

PR #4600, opened 2022-04-16
Sonarr sync: stop using seasonpass API

sephrat:
Follow-up of https://github.com/Ombi-app/Ombi/pull/4589
Fixes https://github.com/Ombi-app/Ombi/issues/4588

Since seasonpass endpoint isn't reliable for us, I suggest to re-use the individual episode logic when the full season is requested and use the series endpoint. 
In that case we will unmonitor every episode of the season, then monitor them again. I agree it's not ideal. Maybe we could optimize this but I was not sure why a similar logic about previously monitored episodes had been commented out.

### DIFF
--- a/src/Ombi.Core/Senders/TvSender.cs
+++ b/src/Ombi.Core/Senders/TvSender.cs
@@ -343,8 +343,6 @@ namespace Ombi.Core.Senders
                 await Task.Delay(500);
             }
 
-            var seriesChanges = false;
-
             foreach (var season in model.SeasonRequests)
             {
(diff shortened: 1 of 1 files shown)


tidusjar:
I think I switched to the season pass as I believed at the time it was more reliable. 

I’ll do some testing! Thanks again for your effort on this, really appreciated

sephrat:
FYI I tested with former versions of Sonarr (V2 and V3) and I had consistent results with the unaired episodes not being monitored with the seasonpass endpoint.

PR #4613, opened 2022-04-19
Hide subscribe button when request is closed

sephrat:
The subscribe button currently still shows when the request is available or denied. 
This PR will hide the button when the request has reached the end of its livecycle.

I anticipated the fix on TV and albums even though I couldn't find that option available in the web app. I'm assuming it wasn't ported to the new V4 layout?

### DIFF
--- a/src/Ombi.Core/Engine/MovieRequestEngine.cs
+++ b/src/Ombi.Core/Engine/MovieRequestEngine.cs
@@ -533,7 +533,7 @@ namespace Ombi.Core.Engine
                 }
                 else
                 {
-                    if (!x.Available && !x.Available4K)
+                    if (!x.Available && !x.Available4K && (!x.Denied ?? true) && (!x.Denied4K ?? true))
                     {
                         x.ShowSubscribe = true;
                     }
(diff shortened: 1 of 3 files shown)


src/Ombi.Core/Engine/MovieRequestEngine.cs
```diff
@@ -533,7 +533,10 @@ private async Task CheckForSubscription(string UserId, List<MovieRequests> movie
                 }
                 else
                 {
-                    x.ShowSubscribe = true;
+                    if (!x.Available && !x.Available4K && (!x.Denied ?? true) && (!x.Denied4K ?? true))
```
tidusjar, line 536: If 4k feature is not enabled, this will never be true as `Available4K` will always be false.

Probably want to do something like this?

```
if ((!x.Available || !x.Available4K) && ((!x.Denied ?? true) || (!x.Denied4K ?? true)))
```

sephrat: > If 4k feature is not enabled, this will never be true as `Available4K` will always be false.

Yes so !x.Available4K will be true and all is fine, no? I've found it to be quite accurate on a non 4k-enabled setups during my tests. ShowSubscribe will be true when the movie is not available at all, false if it's available in any quality. 

However now that you point it out, it may not behave as expected if a 4K request and a regular one is made and only one of those is closed. I'll look into it.

PR #4614, opened 2022-04-19
Add Title to Partially Available Message

Teifun2:
If the Title of the show is not menitoned it can be unclear what episodes are now available.


tidusjar:
Thanks. Keep in mind that this is for settings the default. After it’s been set you can modify it in the settings

Teifun2:
Oh I was not aware of that :D but great to know thanks!

PR #4619, opened 2022-04-23
Add original language filter

sephrat:
image
image

The result is not as good as I expected, looks like some TMDB endpoints we use do not support the new filter (popular does, trending does not). I'll investigate a bit more before submitting this draft.


sephrat:
After more testing I can say the filter does work on all sections of the discovery tab since they all use the same TMDB endpoint. The results were not as satisfying as I thought because I had many filters that only yielded poor quality content. 
Up for testing!

PR #4624, opened 2022-04-25
Improve trending section

sephrat:
It's possible to revert to the previous behavior for now:
image

Movies before:
image

Movies now:
image

TV before:
image

TV now:
image

The different is huge with TV. Movies are debatable.

### DIFF
--- a/src/Ombi.Settings/Settings/Models/FeatureSettings.cs
+++ b/src/Ombi.Settings/Settings/Models/FeatureSettings.cs
@@ -20,6 +20,6 @@ namespace Ombi.Settings.Settings.Models
     public static class FeatureNames
     {
         public const string Movie4KRequests = nameof(Movie4KRequests);
-        public const string NewTrendingSource = nameof(NewTrendingSource);
+        public const string OldTrendingSource = nameof(OldTrendingSource);
     }
 }
--- a/src/Ombi.Settings/Settings/Models/FeatureSettings.cs
+++ b/src/Ombi.Settings/Settings/Models/FeatureSettings.cs
@@ -20,5 +20,6 @@ namespace Ombi.Settings.Settings.Models
     public static class FeatureNames
     {
         public const string Movie4KRequests = nameof(Movie4KRequests);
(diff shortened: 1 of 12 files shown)


src/Ombi/ClientApp/src/app/discover/components/carousel-list/carousel-list.component.ts
```diff
@@ -224,7 +226,11 @@ export class CarouselListComponent implements OnInit {
                 this.movies = await this.searchService.popularMoviesByPage(this.currentlyLoaded, this.amountToLoad);
                 break;
             case DiscoverType.Trending:
+              if(this.isNewTrendingSourceEnabled) {
```
tidusjar, line 229: it looks like we are doing the same thing but two different ways.

For Tv shows it's done in the TvEngine, but for Movies it's done in the UI layer? 

Can we move the Movie implementation to the backend?

src/Ombi.Settings/Settings/Models/FeatureSettings.cs
```diff
@@ -20,5 +20,6 @@ public class FeatureEnablement
     public static class FeatureNames
     {
         public const string Movie4KRequests = nameof(Movie4KRequests);
+        public const string NewTrendingSource = nameof(NewTrendingSource);
```
tidusjar, line 23: I think we should have the inverse, so people get it by default

```
public const string OldTrendingSource = nameof(OldTrendingSource );
```

src/Ombi/ClientApp/src/app/discover/components/carousel-list/carousel-list.component.ts
sephrat: Yeah I initially wanted to build new endpoints for both but when I got to TV I had no choice but to 'break' the existing endpoint.
It didn't feel silly at that time but now that you point it out, I'll make it more consistent.

sephrat:
I made some changes as requested.
I also updated my OP with more screenshots.

PR #4625, opened 2022-04-26
perf: stop populating obsolete subscribe fields from search endpoint

sephrat:
Follow up of https://github.com/Ombi-app/Ombi/pull/4585
API consumers of Subscribed and ShowSubscribe fields should migrate from /search/movie/* endpoint to /Request/movie/info/*.
For the record, these fields could not be trusted because they were cached on a global level whereas these fields are user-dependent. The Request endpoint is not cached and those fields have been fixed for movies in https://github.com/Ombi-app/Ombi/pull/4585.


tidusjar:
This is a kick for me to change the app...

tidusjar:
Change in the app has been made, need to wait for next app release

PR #4630, opened 2022-04-28
feat(discover): Show more relevant shows in upcoming TV

sephrat:
Before:
image
After:
image

Upcoming TV would use filter release_date which does not exist for TVs so this basically turned out to be exactly the same as the popular section. I'm surprised this went unnoticed for so long (by myself included!).
We now use the first air date. I fiddled around and found the sweet spot to be shows that will release in the next month, as opposed to upcoming movies which looks for movies released between D+7 and D+27. 

Now that's a quick win 😃


tidusjar:
Nice work once again!

PR #4633, opened 2022-04-30
Fixed carousel touch not working when scrolling page and recommendations and similar movie navigation

dr3am37:
Simple one line addition to disable the Carousel's component onTouchMove internal function's default behavior.

**Problem**: Carousel element blocks vertical touch input, preventing scrolling the page.

**Info**: By default the carousel would block the user's touch input if trying to scroll the page, making the experience on mobile not fun to use as 80% of the screen is the carousel, so you have to swipe in between them, disabling the default behavior retains the ability to swipe left and right to move forward and back in the carousels but also allows to scroll down the page with no interference to the user.


dr3am37:
New additions since posting this:

Fixed recommendations and similar not changing the data on the component when items are pressed by calling the init function again on param change

Moved the ngif results > 0 to the mat-expansion panel to avoid rendering the entire element if it doesn't have any results instead of having an empty panel.

PR #4637, opened 2022-05-05
Missing Poster broken link fix

dr3am37:
added checks and string formatting in the case of having a missing poster on tmdb to both tv and movies, removed the hard coded tmdb url from the html template and moved into component.


tidusjar:
Thanks!

PR #4648, opened 2022-05-09
Fixes SickRageEpisodeSetStatus model to accept array of objects via data prop

echel0n:
episode results returned from SiCKRAGE API when changing status of episodes is an array of objects, this merge request corrects SickRageEpisodeSetStatus model data property to accommodate that


tidusjar:
Thanks!

PR #4678, opened 2022-06-18
Override Sonarr V3 Profiles endpoint

comigor:
Although the official Sonarr API docs [still states](https://github.com/Sonarr/Sonarr/wiki/Profile) that the Profiles endpoint is `/profile`, that's severely outdated, and the entry on [WikiArr](https://wiki.servarr.com/sonarr) states that "the best way to view see the API is to watch the network calls the frontend makes via the browser development network call console".

Well, Sonarr's [own frontend implementation](https://github.com/Sonarr/Sonarr/blob/93fc9abae91aa4d0093b0df0351d2324b02f8b96/frontend/src/Store/Actions/Settings/qualityProfiles.js#L71) and [NzbDrone proxy](https://github.com/Sonarr/Sonarr/blob/41a821352e3e3e91d5f85cd21da48bd5d0d86452/src/NzbDrone.Core/ImportLists/Sonarr/SonarrV3Proxy.cs#L39) use `/qualityprofile` for it.


comigor:
I'm not too familiar with C#, but this code should be overriding the [GetProfiles](https://github.com/Ombi-app/Ombi/blob/c80c3486c3842432e166614ae79fe47f4c19e073/src/Ombi.Api.Sonarr/SonarrApi.cs#L22) function on the V1 API, right? If so, the change should be transparent unless some test is mocking or expecting a different output

comigor:
It seems automation tests failed with an error unrelated to my changes:
```
AssertionError: expected '2015-01-13T00:00:00' to equal '2015-01-14T01:00:00+00:00'
+ expected- actual
-'2015-01-13T00:00:00'
+'2015-01-14T01:00:00+00:00'
```

tidusjar:
I'll just quickly check over the automation, because we heavily rely on data coming from different providers, when they change the data sometimes it does break the tests :)

tidusjar:
@Mergifyio update

PR #4745, opened 2022-08-30
Fix,Refactor: Fix NullReferenceException by checking default and null values

GermanAizek:
Hi @tidusjar,
this PR will greatly help fix NullReferenceException by checking default and null values and using `?.` operator.
Review and feedback me.


tidusjar:
Hello,

Thanks for the contribution but this doesn't really add much. You might be fixing a potential NullReferenceException but then it will continue to fail further up the stack due to `.?` providing the `default` value.

There are also changes made where it makes no sense to use the null conditional operator because there's a null check just before it.

I'm going to currently deny merging this pull request as it's not targeted enough and could cause unexpected behaviour.

PR #4751, opened 2022-09-07
Add more curly variables for partially available notification

sephrat:
This adds two new variables for partially available notifications:
- `PartiallyAvailableEpisodeCount`: the number of episodes that became available
- `PartiallyAvailableEpisodesList`: the list of episodes that became available with the format `SxE` where S is the season number, and E is the episode number. Example: `1x15`

The first one is a nice to have.
The second one solves an issue with `PartiallyAvailableSeasonNumber` and `PartiallyAvailableEpisodeNumbers` not being clear enough if the notification was about multiple seasons.

This turns the default notification from:

> Season 1 Episodes 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 1, 2

to:

> Episodes 1x1, 1x2, 1x3, 1x4, 1x5, 1x6, 1x7, 1x8, 1x9, 1x10, 2x1, 2x2, 2x3, 2x4, 2x5, 2x6, 2x7, 2x8, 2x9, 2x10, 3x1, 3x2

### DIFF
--- a/src/Ombi.Schedule/Jobs/Plex/PlexAvailabilityChecker.cs
+++ b/src/Ombi.Schedule/Jobs/Plex/PlexAvailabilityChecker.cs
@@ -174,6 +174,8 @@ namespace Ombi.Schedule.Jobs.Plex
                     };
                     notification.Substitutes.Add("Season", availableEpisode.First().SeasonNumber.ToString());
                     notification.Substitutes.Add("Episodes", string.Join(", " ,availableEpisode.Select(x => x.EpisodeNumber)));
+                    notification.Substitutes.Add("EpisodesCount", $"{availableEpisode.Count}");
+                    notification.Substitutes.Add("SeasonEpisodes", string.Join(", ", availableEpisode.Select(x => $"{x.SeasonNumber}x{x.EpisodeNumber}" )));
                     await _notificationService.Notify(notification);
                 }
             }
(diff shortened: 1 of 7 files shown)


src/Ombi.Schedule/Jobs/Plex/PlexAvailabilityChecker.cs
```diff
@@ -174,6 +174,8 @@ private async Task ProcessTv(List<ChildRequests> tv)
                     };
                     notification.Substitutes.Add("Season", availableEpisode.First().SeasonNumber.ToString());
                     notification.Substitutes.Add("Episodes", string.Join(", " ,availableEpisode.Select(x => x.EpisodeNumber)));
+                    notification.Substitutes.Add("EpisodesCount", $"{availableEpisode.Count}");
+                    notification.Substitutes.Add("SeasonEpisodes", string.Join(", ", availableEpisode.Select(x => $"{x.SeasonNumber}x{x.EpisodeNumber}" )));
```
tidusjar, line 178: It would be worth encapsulating this logic to reuse in all of the different places

sephrat: I can't agree more, but the scope is larger than those two lines of code. I think all this block is worth encapsulating, if not more: https://github.com/sephrat/Ombi/blob/partially-available-notification/src/Ombi.Schedule/Jobs/Plex/PlexAvailabilityChecker.cs#L147-L180  
Would you like me to work on it in this PR or should we separate this into another PR?

tidusjar: Yeah that's fair! 

Let's do that in a separate PR, in that case, to keep things small and easier to review!

PR #4767, opened 2022-10-03
Fixes default image for recently requested items.

### LINKED ISSUE #4730: Recently requested image doesn't fallback to default image when no poster found on discover page
**Describe the bug**
I have a film that Fanart TV has no image for (922705).
During the page loading, that film has a nice placeholder image and background.
image

After Ombi returned that no image is found, the placeholder image is removed and the background is black:
image

On a side note, can't we make it so that the poster defaults to the one from TMDB if not found from Fanart TV?

bernarden:
Fixes #4730.

Before:
image

After:
image


tidusjar:
nice work

PR #4769, opened 2022-10-05
Upgrades nuget packages. Removes deprecated packages. Fixes build warnings.

bernarden:
This PR upgrades all nuget packages and removes usages of obsolete methods. 
The only 4 warning remaining are due to https://github.com/AutoMapper/AutoMapper.Collection/pull/171 not being released yet.

Notable changes:
MusicBrainzClient client is used instead of calling obsolete static methods.
Microsoft.Extensions.PlatformAbstractions is removed and the code that retrieves a runtime version has changed but should return identical result.
Microsoft.AspNetCore.SignalR is consistently used across the solution instead of a mix of Microsoft.AspNet.SignalR and Microsoft.AspNetCore.SignalR references.
INotificationHubService provides an abstraction from SignalR. NotificationHubService wraps NotificationHub and exposes all required functionality.
GenreIds property on MovieDbSearchResult model changed from int?[] to int[]. This was done due to AutoMapper error. I didnt see any existing usages of that property or a reason why it needs to be int?[].
Serilog.Sinks.RollingFile is removed. Same rolling log functionality now exists in Serilog.Sinks.File package.

### DIFF
--- a/src/Ombi.Api.Mattermost/Ombi.Api.Mattermost.csproj
+++ b/src/Ombi.Api.Mattermost/Ombi.Api.Mattermost.csproj
@@ -11,7 +11,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
   </ItemGroup>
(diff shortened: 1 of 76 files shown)


tidusjar:
Excellent work here! Really nice job!

Issue #4688, opened 2022-07-18
Provide a flag for missing users on Plex Server


ChrisColotti:
Was asked to add here from Discord channel but...

While Ombi tries to remove stale users I've seen sometimes they remain and have to do manual cleanup.  Tautulli solves this with a simple yellow flag that states "User not found on plex Server" when it does a user API pull.  This can help speed up user Management when you clean up sharing on your plex account or a user removes themselves so Ombi is always up to date with the same users even if needed to be done manually.  The stale users will be easy to find.

ChrisColotti:
@tidusjar I did add as an official "feature" as well since this had been auto closed LOL

ChrisColotti:
I do know it's not cleaning out missing users I tested removing someone the other day.  Tautulli shows them as missing, but the same user is still listed in the Ombi DB.  This can definitely use a little tweaking to either automatically or manually keep things lined up.  Thanks for everything!

sephrat:
Added in v4.29.0

PR #4778, opened 2022-10-15
Provide a flag for missing users on Plex Server (#4688)

xweskingx:
* Add opt-in feature to delete users that no longer exist in Plex when configured for Import Plex Users or Import Plex Admin


tidusjar:
Looks good, nice work

PR #4787, opened 2022-10-26
Only log error messages from Microsoft

sephrat:
This will remove warnings that are only meant for developers and are misleading for users.


tidusjar:
I have been reviewing them but you are ignoring some https://github.com/Ombi-app/Ombi/pull/4776

PR #4841, opened 2023-01-03
Add the option for header authentication to create users

janderedev:
Since I manage my users externally using authentik, manually adding every user to Ombi is not feasable for me.

This PR adds a toggle to the authentication settings, and if enabled, a new user will be created if no existing user can be found when header authentication is being used. The default user settings defined in the User Management tab (roles, request limit, streaming country) will be applied.

image

### DIFF
--- a/src/Ombi.Settings/Settings/Models/AuthenticationSettings.cs
+++ b/src/Ombi.Settings/Settings/Models/AuthenticationSettings.cs
@@ -15,5 +15,6 @@ namespace Ombi.Settings.Settings.Models
         public bool EnableOAuth { get; set; } // Plex OAuth
         public bool EnableHeaderAuth { get; set; } // Header SSO
         public string HeaderAuthVariable { get; set; } // Header SSO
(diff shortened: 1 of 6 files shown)


tidusjar:
Looks good!

Nice contribution!


There's only one thing that i'd request to add, just some text below the slider indicating that it could be misused if someone calls the header auth API, they would be able to create a user on your Ombi account, so ensuring that the default roles being set would be quite important.

janderedev:
@tidusjar Thanks! Ideally I would add 2 warnings like this:

image

Is that fine?

tidusjar:
Yeah excellent!

janderedev:
All right, pushed!

If this gets merged, I can also open a pull request to the Docs repository to document the new setting.

tidusjar:
Sounds perfect!

janderedev:
Thanks for merging!

I created a pull request (Ombi-app/Ombi.Docs#222) to the Docs repo :)

PR #4912, opened 2023-04-24
Search by genre

ryan-c44:
From discovery page can now view a list of buttons that represent each genre.  When user clicks on a genre button they will be taken to a page of movies within that genre.


tidusjar:
This is great! Nice work! I'll merge it in, I want to just place some margin around the buttons but good job :)

PR #4914, opened 2023-04-27
Feature/search by genres

ryan-c44:
Include radio select button to toggle between movie/tv type.


tidusjar:
There's a few conflicts, do you mind resolving them?

Issue #4915, opened 2023-04-27
Can't request on behalf with the API. 


phildups7:
**Describe the bug**
When doing a request through the API with a value for requestOnBehalf I get this error "You do not have the correct permissions to request on behalf of users!". If I don't put any values for requestOnBehalf the requests goes through. Same issue also happens in swagger. 
The API key doesn't give the privilege to request on behalf

**To Reproduce**
1. Go to swagger
2. Set the API key
3. Go to Post /api/v1/Request/movie or /api/v2/Requests/tv
4. Click on Try it out
5. Make sure there's a value for requestOnBehalf
6. Click on Execute
7. See error

**Expected behavior**
The request should go though.
In Ombi, under request, the Requested By should show the value entered on step 5

**Screenshots**
I can add screenshots if needed, but should be easy to reproduce with swagger.

**Logs (Logs directory where Ombi is located)**
{
  result: false,
  message: 'You do not have the correct permissions to request on behalf of users!',
  isError: true,
  errorMessage: 'You do not have the correct permissions to request on behalf of users!',
  errorCode: null,
  requestId: 0
}.

**Desktop (please complete the following information):**
 - OS: Windows 10 

**Ombi Version (please complete the following information):**
 - Version 4.35.10
- Media Server: Plex but not used in this case 
- Database Type: SQLite 

**Additional context**

PR #4919, opened 2023-05-02
Bugfix: Allow RequestOnBehalf if requested from the API

phildups7:
This is a fix for a bug I found with the RequestOnBehalf. https://github.com/Ombi-app/Ombi/issues/4915
Issue 4915 - Can't request on behalf with the API. 

Username.Equals("API", StringComparison.CurrentCultureIgnoreCase) is used in BaseEngine so I figured it's probably the best way to check for API unless a special role is created in OmbiRoles. I tried to keep the changes as minimal as possible but could rework this if needed.

This only works if requestOnBehalf value is a valid user Id or the API user id. This fix mirrors the behavior of the UI where you also need to select an existing user.

### DIFF
--- a/src/Ombi.Core/Engine/MovieRequestEngine.cs
+++ b/src/Ombi.Core/Engine/MovieRequestEngine.cs
@@ -77,7 +77,8 @@ namespace Ombi.Core.Engine
             var userDetails = await GetUser();
             var canRequestOnBehalf = model.RequestOnBehalf.HasValue();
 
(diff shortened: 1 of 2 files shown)


phildups7:
This is a fix for a bug I found with the RequestOnBehalf. https://github.com/Ombi-app/Ombi/issues/4915

`Username.Equals("API", StringComparison.CurrentCultureIgnoreCase)` is used in BaseEngine so I figured it's probably the best way to check for API unless a special role is created in OmbiRoles. I tried to keep the changes as minimal as possible but could rework this if needed.

PR #4922, opened 2023-05-09
Add the RequestedByAlias field to the Notification Message

phildups7:
RequestedByAlias field was missing from the notification message. I was trying to use it part of the webhooks and realized it was missing. 

In the following test I used the web UI to send the request so RequestedByAlias is null:
requestByAliasNull

In the following test I sent the request using the API and set a value for RequestedByAlias in the request header, so RequestedByAlias is not null:
requestByAliasNotNull

I placed it right after alias but could move it anywhere if needed.


tidusjar:
Nice spot, thanks!

PR #4881, opened 2023-03-09
First step towards played sync

sephrat:
This PR is a first step towards syncing played states from media servers to Ombi.
The long-term goal for this PR and future PRs is the following:

- Sync episodes and movies played status for every user, for every media server
- Show this information along with the requests to know if the user who made the request (and possibly other users) has watched it.

The current development progress for this PR is:

- [x] Cache Emby movie played state in Ombi
- [x] Enhance the requests API with "played" state
- [x] Display played state in the requests list
- [x] Allow to toggle this feature on/off (default: off), as this is highly experimental

Out of scope features for this PR (other PRs will follow):

- Episode sync
- Plex and Jellyfin sync

Some technical documentation:

- Refactored EmbyContentSync to reuse parts of code for EmbyPlayedSync
- Created a new table UserMoviePlayed in ExternalDb to cache played state

Let me know what you think of this and my plans!

PR #4928, opened 2023-05-17
Sync episodes played states from Emby

sephrat:
This is the TV equivalent of https://github.com/Ombi-app/Ombi/pull/4881

image
A tooltip on the column header explains its meaning:
image

I'm curious to know if there's a more elegant way to do what I did in src/Ombi.Core/Engine/TvRequestEngine.cs, I'm not sure I'm using the most of C# capabilities here.

Once this PR is merged, I'm planning to work on bringing these features to Jellyfin and Plex sync.

### DIFF
--- a/src/Ombi.Api.Emby/EmbyApi.cs
+++ b/src/Ombi.Api.Emby/EmbyApi.cs
@@ -249,15 +249,11 @@ namespace Ombi.Api.Emby
             req.AddHeader("Device", "Ombi");
         }
 
-        public async Task<EmbyItemContainer<EmbyMovie>> GetMoviesPlayed(string apiKey, string parentIdFilder, int startIndex, int count, string userId, string baseUri)
-        {
-            return await GetPlayed<EmbyMovie>("Movie", apiKey, userId, baseUri, startIndex, count, parentIdFilder, "ProviderIds");
-        }
+        public async Task<EmbyItemContainer<EmbyMovie>> GetMoviesPlayed(string apiKey, string parentIdFilder, int startIndex, int count, string userId, string baseUri) =>
+            await GetPlayed<EmbyMovie>("Movie", apiKey, userId, baseUri, startIndex, count, parentIdFilder, "ProviderIds");
 
-        public async Task<EmbyItemContainer<EmbyEpisodes>> GetTvPlayed(string apiKey, string parentIdFilder, int startIndex, int count, string userId, string baseUri)
-        {
-            return await GetPlayed<EmbyEpisodes>("Episode", apiKey, userId, baseUri, startIndex, count, parentIdFilder);
-        }
+        public async Task<EmbyItemContainer<EmbyEpisodes>> GetTvPlayed(string apiKey, string parentIdFilder, int startIndex, int count, string userId, string baseUri) =>
+           await GetPlayed<EmbyEpisodes>("Episode", apiKey, userId, baseUri, startIndex, count, parentIdFilder);
 
         private async Task<EmbyItemContainer<T>> GetPlayed<T>(
             string type,
--- a/src/Ombi.Api.Emby/EmbyApi.cs
+++ b/src/Ombi.Api.Emby/EmbyApi.cs
@@ -251,16 +251,32 @@ namespace Ombi.Api.Emby
 
         public async Task<EmbyItemContainer<EmbyMovie>> GetMoviesPlayed(string apiKey, string parentIdFilder, int startIndex, int count, string userId, string baseUri)
         {
-            return await GetPlayed<EmbyMovie>("Movie", apiKey, userId, baseUri, startIndex, count, parentIdFilder);
+            return await GetPlayed<EmbyMovie>("Movie", apiKey, userId, baseUri, startIndex, count, parentIdFilder, "ProviderIds");
         }
 
-        private async Task<EmbyItemContainer<T>> GetPlayed<T>(string type, string apiKey, string userId, string baseUri, int startIndex, int count, string parentIdFilder = default)
+        public async Task<EmbyItemContainer<EmbyEpisodes>> GetTvPlayed(string apiKey, string parentIdFilder, int startIndex, int count, string userId, string baseUri)
+        {
+            return await GetPlayed<EmbyEpisodes>("Episode", apiKey, userId, baseUri, startIndex, count, parentIdFilder);
+        }
+
+        private async Task<EmbyItemContainer<T>> GetPlayed<T>(
+            string type,
+            string apiKey,
+            string userId,
+            string baseUri,
+            int startIndex,
+            int count,
+            string parentIdFilder = default,
+            string fields = default)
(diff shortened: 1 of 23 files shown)


src/Ombi.Core/Engine/TvRequestEngine.cs
```diff
@@ -896,6 +905,52 @@ private async Task CheckForSubscription(HideResult shouldHide, List<ChildRequest
...
+                var requestedEpisodes = getEpisodesKeys(request);
+
+                var playedEpisodes = _userPlayedEpisodeRepository
+                    .GetAll()
+                    .Where(x => x.TheMovieDbId == request.Id && x.UserId == request.RequestedUserId)
+                    .AsEnumerable()
+                    .Join(requestedEpisodes,
+                        played => new { played.SeasonNumber, played.EpisodeNumber },
+                        requested => new { requested.SeasonNumber, requested.EpisodeNumber },
+                        (played, requested) => new { played });
```
tidusjar, line 928: The above, in the join we are not using the result from `requested`, so you can remove that

src/Ombi.Core/Engine/TvRequestEngine.cs
```diff
@@ -896,6 +905,52 @@ private async Task CheckForSubscription(HideResult shouldHide, List<ChildRequest
...
+
+                var playedCount = playedEpisodes.Count();
+                var toWatchCount = requestedEpisodes.Count();
+                request.RequestedUserPlayedProgress = 100 * playedCount / toWatchCount;
+                
+            }
+        }
+
+        private List<EpisodeKey> getEpisodesKeys(ChildRequests request)
```
tidusjar, line 937:
```suggestion
        private List<EpisodeKey> GetEpisodesKeys(ChildRequests request)
```

src/Ombi.Api.Emby/EmbyApi.cs
```diff
@@ -251,16 +251,32 @@ private static void AddHeaders(Request req, string apiKey)
 
         public async Task<EmbyItemContainer<EmbyMovie>> GetMoviesPlayed(string apiKey, string parentIdFilder, int startIndex, int count, string userId, string baseUri)
         {
-            return await GetPlayed<EmbyMovie>("Movie", apiKey, userId, baseUri, startIndex, count, parentIdFilder);
+            return await GetPlayed<EmbyMovie>("Movie", apiKey, userId, baseUri, startIndex, count, parentIdFilder, "ProviderIds");
         }
 
-        private async Task<EmbyItemContainer<T>> GetPlayed<T>(string type, string apiKey, string userId, string baseUri, int startIndex, int count, string parentIdFilder = default)
+        public async Task<EmbyItemContainer<EmbyEpisodes>> GetTvPlayed(string apiKey, string parentIdFilder, int startIndex, int count, string userId, string baseUri)
```
tidusjar, line 257: Can be simplified

```
public Task<EmbyItemContainer<EmbyEpisodes>> GetTvPlayed(string apiKey, string parentIdFilder, int startIndex, int count, string userId, string baseUri) =>
 GetPlayed<EmbyEpisodes>("Episode", apiKey, userId, baseUri, startIndex, count, parentIdFilder);


```

src/Ombi.Schedule/Jobs/Emby/EmbyPlayedSync.cs
```diff
@@ -98,13 +113,117 @@ private async Task ProcessMovie(EmbyMovie movieInfo, OmbiUser user, ICollection<
...
+
+        private async Task ProcessTvUser(EmbyServers server, OmbiUser user, string parentId = default)
+        {
+            EmbyItemContainer<EmbyEpisodes> episodes;
+            if (recentlyAdded)
+            {
+                var recentlyAddedAmountToTake = 10; // to be adjusted?
+                episodes = await Api.GetTvPlayed(server.ApiKey, parentId, 0, recentlyAddedAmountToTake, user.ProviderUserId, server.FullUri);
+                // Setting this so we don't attempt to grab more than we need
+                if (episodes.TotalRecordCount > recentlyAddedAmountToTake)
```
tidusjar, line 132:
```suggestion
                if (episodes.TotalRecordCount >= recentlyAddedAmountToTake)
```

src/Ombi.Schedule/Jobs/Emby/EmbyPlayedSync.cs
```diff
@@ -98,13 +113,117 @@ private async Task ProcessMovie(EmbyMovie movieInfo, OmbiUser user, ICollection<
...
+                UserId = user.Id
+            });
+
+            if (episode.IndexNumberEnd.HasValue && episode.IndexNumberEnd.Value != episode.IndexNumber)
+            {
+                int episodeNumber = episode.IndexNumber;
+                do
+                {
+                    _logger.LogDebug($"Multiple-episode file detected. Adding episode ${episodeNumber}");
+                    episodeNumber++;
```
tidusjar, line 195: This being in a `do while` loop, the first episodeNumber that get's saved is going to be incremented. Is this expected?

src/Ombi.Schedule/Jobs/Emby/EmbyPlayedSync.cs
sephrat: If the two variables are equal, it's not necessary the assign one to the other is it?
For the record, this is a copy and paste from the episode sync (from https://github.com/sephrat/Ombi/blob/9fe7ec115e8657ed77028cbdfce1e449bfb22c1f/src/Ombi.Schedule/Jobs/Emby/EmbyEpisodeSync.cs#L112).

src/Ombi.Schedule/Jobs/Emby/EmbyPlayedSync.cs
sephrat: The first episode has already been added a few instructions before. This will add all episodes after this one when the file spans over multiple episodes.
FYI, this is a copy and paste from the episode sync: https://github.com/sephrat/Ombi/blob/9fe7ec115e8657ed77028cbdfce1e449bfb22c1f/src/Ombi.Schedule/Jobs/Emby/EmbyEpisodeSync.cs#L171-L174

src/Ombi.Core/Engine/TvRequestEngine.cs
sephrat: I'm not sure what you mean but this is probably because I'm still discovering the Join feature.
Do you mean this?
```
.Join(requestedEpisodes,
                        played => new { played.SeasonNumber, played.EpisodeNumber },
                        requested => new { requested.SeasonNumber, requested.EpisodeNumber },
                        (played) => new { played });
```
If so, this gives me a syntax error (CS0411). Am I missing something?

If you mean that the join itself is not necessary, I use it to additionally filter the result of my query so that it only returns episodes that are in requestedEpisodes. This is what I tried to explain on Discord earlier.

Issue #5046, opened 2023-12-07
Landing Page with Text Causes Continue Button to not be visible.


OcifferAction:
### Summary

I only have a few sentences on my landing page, but on smaller phones it blocks out the continue button for my users to log in. I'm unable to scroll any further down. On my higher resolution Galaxy 21 I can see the continue button, but I have to keep the message very short. I've attached a screenshot from my phone.
[https://imgur.com/UJV6woK](https://imgur.com/UJV6woK)

### Ombi Version

4.43.5

### What platform(s) does this occur on?

Linux

### What database are you using?

SQLite (Default)

ExtremeFiretop:
I can confirm this is still an issue with the current version of Ombi, I just noticed this on Ombi 
4.43.21 for Windows.

ExtremeFiretop:
@OcifferAction 

This is fixed as of version 4.43.22

PR #5127, opened 2024-07-08
Update landingpage.component.html To Page Scrolling.

ExtremeFiretop:
Update landingpage.component.html To Page Scrolling on mobile devices.

This is to address issue:

https://github.com/Ombi-app/Ombi/issues/5046

Tested on my mobile with success.

### DIFF
--- a/src/Ombi/ClientApp/src/app/landingpage/landingpage.component.html
+++ b/src/Ombi/ClientApp/src/app/landingpage/landingpage.component.html
@@ -1,4 +1,4 @@
-﻿<div *ngIf="landingPageSettings && customizationSettings" style="overflow:hidden">
+﻿<div *ngIf="landingPageSettings && customizationSettings" style="overflow: auto !important;">
     <ombi-image-background></ombi-image-background>
 
     <div class="small-middle-container">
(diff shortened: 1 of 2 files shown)


tidusjar (changes requested):
Thanks for the PR!

The only problem with this is that due to some other content div it seems there's always a scrollbar no matter the viewport size now. Looks like the `row` in the `small-middle-container` is causing it oddly enough

ExtremeFiretop:
> Thanks for the PR!
> 
> The only problem with this is that due to some other content div it seems there's always a scrollbar no matter the viewport size now. Looks like the `row` in the `small-middle-container` is causing it oddly enough

I'm sorry if I over-stepped. I was just trying to learn a bit myself and troubleshoot at the same time.
I did mention this was only a workaround in the issue. So if this is not correct please feel free to close.

Otherwise, I can keep poking?

tidusjar:
It's completely fine! Please feel free to keep poking away :)

ExtremeFiretop:
> Thanks for the PR!
> 
> The only problem with this is that due to some other content div it seems there's always a scrollbar no matter the viewport size now. Looks like the `row` in the `small-middle-container` is causing it oddly enough

@tidusjar 

I believe I fixed it in commit: https://github.com/Ombi-app/Ombi/pull/5127/commits/18e0d90f49d3b3b64c8bea2a31daa7ce68c90fd4

Can you test/review and advise?

ExtremeFiretop:
On my desktop with a ultrawide screen at 1440p, I have no scroll bar.
On my phone, I can scroll down to the button even with long text at the landing page.

(I am actually using HTML for the landing page, still no issues though)

tidusjar:
Nice work!

tidusjar (approved):

Issue #4994, opened 2023-08-02
Can't submit Radarr settings. `normalForm` undefined. 


steezeburger:
### Summary

`normalForm` is undefined on the Radarr settings page. `Submit` button always greyed out. Can't change settings. Nothing relevant in the logs that I could see.

The first time I set the Radarr settings, I did try testing the connection but never got any feedback, then I clicked submit. So I think I submitted in the middle of a pending request. Did this throw off the app state in a way that causes the `normalForm` value to be undefined on subsequent loads?

https://github.com/Ombi-app/Ombi/blob/500b65ea7e67abb8a184f21de7c0c37ebafe8f3c/src/Ombi/ClientApp/src/app/settings/radarr/radarr.component.ts#L69

<img width="1047" alt="Screenshot 2023-08-02 at 2 39 22 AM" src="https://github.com/Ombi-app/Ombi/assets/5599073/8622350a-21d4-42b2-95d5-c5e93efec046">

### Ombi Version

4.39.1

### What platform(s) does this occur on?

Linux

### What database are you using?

SQLite (Default)

### Relevant log output (browser console)

```shell
main.2f123838503b8dd3.js:11 ERROR TypeError: Cannot read properties of undefined (reading 'changes')
    at Object.next (943.114b5c43c90542ed.js:1:371057)
    at q.next (main.2f123838503b8dd3.js:11:501692)
    at le._next (main.2f123838503b8dd3.js:11:501371)
    at le.next (main.2f123838503b8dd3.js:11:501062)
    at main.2f123838503b8dd3.js:1:285142
    at Ge.runInsideAngular (main.2f123838503b8dd3.js:1:282413)
    at Ge.leave (main.2f123838503b8dd3.js:1:282340)
    at Ge.leave (main.2f123838503b8dd3.js:1:285391)
    at Object.next (main.2f123838503b8dd3.js:1:285129)
    at q.next (main.2f123838503b8dd3.js:11:501692)
```

JakeMHughes:
I also have this issue on version 4.43.8

steezeburger:
@JakeMHughes I unfortunately had to uninstall and reinstall Ombi and was very careful with how I input the Radarr settings, but I was able to get it working.

I actually ran into this issue again recently when setting up a new server. I had to uninstall/reinstall once again.

I can't remember my exact steps, but I believe I:
1. Input Radarr server configuration settings, but did not fill out Radarr interface settings yet
2. Click test connectivity button
[...]

PR #5207, opened 2025-01-03
fix(radarr-settings): this.normalForm is undefined

AlexandrePicavet:
Hi I fixed an error about the `normalForm` in the Radarr Settings, this is my first contribution here, I hope this PR meet the standards :slightly_smiling_face:.

This error happens because the `normalForm` component is nested within an `ngIf`, waiting for the `form` variable to be initialized.
In the `ngOnInit` method, when the `radarrFacade.state$()` emits a value, we synchronously initialize the form and try to access the `changes` field of the `normalForm` reference one after the other, but the `normalForm` remains `undefined` since the template has not yet been updated.

Fixes Ombi-app/Ombi#4994


tidusjar:
Thanks! I'll take a look :)

tidusjar:
Looks good :)

PR #5262, opened 2025-09-11
docker for local build

devildant:
I created a Docker setup to be able to develop on Ombi locally without affecting the system. I’m not sure if this might be of interest to you, so just in case I’m sharing it here.


devildant:
Hi @tidusjar,

I also rebased this PR, I don't know if it will interest you, but I found it more practical for dev to have a docker

PR #5263, opened 2025-09-11
filter server plex

devildant:
As a follow-up to my previous PR : https://github.com/Ombi-app/Ombi/pull/5261

i add filter for show only ther server in the list (see my last comment in my previous PR)


tidusjar:
Nice!

devildant:
Hi, 
I rebased

